PR #129, opened 2015-03-16
remove ng animate

domoritz:


domoritz:
@kanitw should we get this in?

PR #184, opened 2015-09-23
NPM3 compatibility and NPM script aliases for Gulp tasks

kadamwhite:
- Resolve missing peer dependencies that cause errors with NPM3
- Add npm script aliases for gulp commands
- Auto-install bower packages on `npm install` using a postinstall script


domoritz:
@kadamwhite This is great! Can you send a pr for https://github.com/vega/polestar as well?

PR #186, opened 2015-09-24
Fix issue where display area does not take up full height

kadamwhite:
See vega/vega-lite-ui#73 and vega/polestar#322

This was introduced in dea6557d32c7733d220bd5dfca7bcc62f22b4dc8 while addressing an issue that prevented the bookmarks from closing: the display height issue arose due to the introduction of another layer of nesting that the CSS did not account for.

The same angular hierarchy can be achieved with less DOM nesting, so this commit addresses the overflow by removing unneeded intermediate div nodes.

Note that this did not end up being an issue with vega-lite-ui, so this closes vega/vega-lite-ui#73 as a dupe of an issue that exists separately within polestar and voyager.

Polestar commit to follow.

### DIFF
--- a/src/app/main/main.html
+++ b/src/app/main/main.html
@@ -1,5 +1,4 @@
-<div ng-controller="MainCtrl">
-  <div class="flex-root vflex full-width full-height">
+<div ng-controller="MainCtrl" class="flex-root vflex full-width full-height">
     <div class="full-width no-shrink">
       <div class="card no-right-margin no-top-margin margin-bottom">
         <div class="right flex-root hflex">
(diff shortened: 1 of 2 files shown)


src/app/main/main.html
```diff
@@ -1,5 +1,4 @@
-<div ng-controller="MainCtrl">
-  <div class="flex-root vflex full-width full-height">
+<div ng-controller="MainCtrl" class="flex-root vflex full-width full-height">
     <div class="full-width no-shrink">
```
kadamwhite: Once I get a go-ahead to merge on this PR I will add a whitespace commit that resolves the indentation inconsistency—I left it this way to make reviewing easier

kadamwhite:
@domoritz @kanitw this and vega/polestar#327 are ready for review

PR #190, opened 2015-09-30
Add GA Logging to vlui

vlandham:
Companion PR to https://github.com/vega/vega-lite-ui/pull/83 . Google analytics is now incorporated in vlui - so can be removed from voyager.

Also adds a few more events for data loading


domoritz:
@vlandham can you add this to polestar as well?

Issue #191, opened 2015-09-30
UI Refresh Mockups & Discussion


vlandham:
Working with @iamjessklein , we have developed a few mockups around suggestions for where we could take the voyager UI. Our primary goals for looking at the UI are:

* Reduce friction for new users coming to voyager
* Provide a stronger visual hierarchy around the different components used in voyager
* Provide more context around the visualizations generated in the tool and what makes voyager unique

### Main View

Here is a rough mockup of what loaded data might look like in Voyager: 

![voyager_with_data](https://cloud.githubusercontent.com/assets/9369/10206086/f98c4706-6779-11e5-9d3c-657c6ad00791.png)

Again - a rough mockup. Anything that looks out of alignment or off is from me hacking on Jess's wonderful mockups. 

Most of the components remain the same. The color changes help to draw the eye over to the left panel where most of the initial interaction for a user will occur. 

In the top bar, we move some of the debugging tools down to the bottom section of the left panel - allowing us to emphasize just the tools that most users will want to use. There is also the concept of a user account here. I think this will be a later addition - and perhaps we will not initially have full names and such - but with a tool built around pulling in custom data sources and bookmarking, having accounts in some form (perhaps initially just github accounts) makes a lot of sense.

In the left panel, we emphasize the data loading process by making it more explicit around how to change the data you are looking at.

In the main panel, we try to be a bit more explanative in what is being shown. We want to also again highlight the hierarchy of this section by distinguishing the different sections here. 

There is also a view toggle in the upper right-hand corner of the main panel - which could transition between a grid and a list based view of the visualizations. I think this could help constrain the main view for new users - which could help when browsing - but i think this will probably come later. 

### Data Panel

Clicking on the "Explore Other Data" link will expand this area, covering or pushing down the 'fields' section of the left panel. It would look something like this:

![voyager_explore_data](https://cloud.githubusercontent.com/assets/9369/10206547/4fa6ab66-677c-11e5-8ac9-91acdc37079d.png)

Here the sample data menu could be used to get another sample data. Or custom data could be uploaded. 

### Data Upload

This is the same as the mockup used for https://github.com/vega/vega-lite-ui/issues/77 - but now integrated with this new look

![voyager_import_data](https://cloud.githubusercontent.com/assets/9369/10206650/d91a8dd6-677c-11e5-93f5-291b75c1eff4.png)

Would love to get thoughts on these changes @kanitw @domoritz @kadamwhite

kanitw:
[...]

PR #195, opened 2015-10-14
Jim/update ui

vlandham:
This is a start of moving towards the UI discussed in #191 . Here is what voyager looks like from this PR:

screen shot 2015-10-14 at 9 50 08 am

You can see I haven't started changing the colors of the panels - but otherwise, most of the suggestions are in place. New logo is there. Links are moved around. Wording is changed for main panel. 

This moves some of shared css from vlui to voyager and polestar respectively. See companion PR:
https://github.com/vega/vega-lite-ui/pull/107

From what I can tell, this CSS is not used directly in vlui - so it makes sense to me to move it closer to the application. As polestar's UI changes or merges it wouldn't be difficult to move it back. 

If this solution isn't acceptable - then i could also work on a way to override shared CSS inside voyager.


src/app/index.scss
```diff
@@ -1,6 +1,7 @@
 // FIXME should use normalize.scss version instead
 // but I got some errors so I don't want to spend too much time here now.
 @import "../assets/normalize.scss";
+@import "../assets/angular-tour.scss";
```
domoritz: doesn't exist yet

vlandham: yep - sorry. just saw that.

vlandham:
ok. here is a revision with the different colors as suggested in the mockups. Thoughts?

screen shot 2015-10-14 at 1 22 14 pm

vlandham:
ping @domoritz & @kanitw

jheer:
Looks good to me! Perhaps we should update the standard link colors (e.g.,
at the top of the page) to match the more purple base tone?

On Wed, Oct 14, 2015 at 1:27 PM, Jim Vallandingham <notifications@github.com

vlandham:
good point @jheer - the mockups have them matching. I will update.

vlandham:
updated with matching link colors

screen shot 2015-10-14 at 1 36 18 pm

src/app/index.scss
```diff
@@ -9,17 +9,121 @@ $fa-font-path: "../../bower_components/fontawesome/fonts";
...
+
+.voyager a {
+  color: #38425D;
+}
+
+.voyager a:hover {
+  color: #72757B;
+}
+
+h1 {
```
kanitw: this doesn't need to be moved.

kanitw:
@vlandham  looks like some of the things you moved are copied to polestar as well.  I think we should only move things that are specific to Voyager.  

The latest stuff looks cool. Here are some nits:

pasted_image_10_14_15__3_23_pm

pasted_image_10_14_15__3_25_pm

Maybe the height of the top bar can be made by adjusting the logo like this?
slice 1

vlandham:
@kanitw i'm with you on everything but the different color backgrounds for the different sections. With grey - we have another 'background' color which is confusing. With white - the section headings bleed into the generated charts. I can try these again and provide some screenshots for what it might look like.

src/app/index.scss
vlandham: as i modified h2 for voyager - i thought it would make sense to just keep all the headers together. I can move them back to vlui though.

vlandham:
And i'm not crazy about smashing the cool new logo! A little space on top isn't too bad right? Lets the user breath a bit.

domoritz:
I'm with Ham on this. We want the top bar to be as small as possible. Especially when we embed voyager/polestar we want every pixel.

vlandham:
i don't disagree that every pixel is important - but sometimes the space around content is just as important right? having too small of a navigation bar reduces the distinction and hierarchy of the bar and thus the rest of the site.  I will work on reducing the size, but do think there is room for a bit of space in the tool.

kanitw:
I agree that too small navigation bar can be bad too, but I still feel that it is currently too big especially that our tool is more like an app more than a website.  

Note that I do not like the size of the IDL logo in the old header bar -- it was too big, but haven't got a chance to fix it.  So feel free to reduce its size too.  


Another question -- Do you plan to implement color change on polestar too?

vlandham:
Ok. no problem. Now the header is 49px tall.

screen shot 2015-10-14 at 7 37 10 pm

If we feel this is the way to go with voyager - i wouldn't be opposed to moving he same stylings to polestar. It would, perhaps, be beneficial to determine if these two tools will be integrated first - which could change the priority of the task.

jheer:
We may want to integrate Polestar functionality into Voyager, but in any
case will want to keep Polestar as a stand alone project!

kanitw:
@vlandham  Thanks for the header change.  I like it more this way.    

I have a few more questions:

1) What happens now when users hover over a field pill?
We used to show yellow highlight on field pill hover like this:

pasted_image_10_14_15__8_01_pm

I want to make sure that the style for this behavior is updated properly ;)


2) Do we really need one line to always display "Load your data..."?

3) Do you have the sketch file for the logo?  (Or it is just in the mock file that you send me?)
I want to fiddle with the logo sizing a bit.

vlandham:
1) right now the same yellow is applied to the hovered pills - as expected. we could definitely tweak the color - but i kept that out of this PR for now.

screen shot 2015-10-14 at 8 10 52 pm

2) the feeling is that this tool is only really useful to folks outside of the user study if they can load and explore their own data. The most direct way to allow people to understand this is by providing a visual cue. I think a small link like this is fairly un-obtrusive and provides a direct link to the load data module - which is what people will want to get to once they understand what the tool can do.

3) the logo was also made in sketch. I'll send you an updated file with it isolated.

kanitw:
Thanks for your reply :) 

1) The color actually still looks okay, so I don't mind having it as is.  Although it could be further tweaked, I agree that it should be in a separate PR. 

2) 
> the feeling is that this tool is only really useful to folks outside of the user study if they can load and explore their own data. The most direct way to allow people to understand this is by providing a visual cue. 

Given your reasoning,

2.1) should the link be in the [welcome screen](https://cloud.githubusercontent.com/assets/9369/10326930/2674d162-6c6f-11e5-8cc0-c63b133b7cbf.png) or be in some form of pointer tooltip (like the [Shepherd](http://github.hubspot.com/shepherd/))?

2.2) Or if you think that it should still be here?  Should the link disappear after it is loaded first time?
Again our future Voyager extension might require some space on the left pane.

3) Awesome Thanks!


4) New Idea: I start thinking that we might be able to fit all the links on the bottom-left corner in one line like this:  
```
About  Debug  Report Bug
```

vlandham:
For 2) - I definitely want to get it into the in-app tutorial - so that it highlights that loading custom data is a possibility. I think that if vertical scrolling is enabled for the left panel, perhaps space will be less of an issue in the future? Else, then yes - perhaps more complex measures could be used to remove it after an initial use. though then it would be difficult for said users to find again!

kanitw:
2)  Vertical scroll is always the backup plan, but we should avoid them if possible as interaction with scrolling could be annoying in general.   

I think the issue that you mentioned can be resolved like this: 
```
Data:  <Dataset_name> \[Load\]
```

So the load button is not obscured, while doesn't take too much space.  

Also, your point about "this tool is only really useful to folks outside of the user study if they can load and explore their own data" is very good.  In fact, the data shouldn't be a select button and the list of examples dataset should be just another "Example dataset" tab in the data load modal.

That said, my suggested solution here requires a lot of changes, so it should be in a separate PR so I'm okay regarding 2) for this PR.  

3) I think I'll send a PR fiddling a bit with header and we can merge this.  :)

src/app/index.scss
```diff
@@ -9,17 +9,104 @@ $fa-font-path: "../../bower_components/fontawesome/fonts";
...
+
+#logo {
+  height: 34px;
+  flex-shrink: 0;
+}
+
+.controls {
+  margin-left: 33px;
+  margin-top: 5px;
+  a {
```
kanitw: `.controls > a` is always `a.command`
better use `& > a.command` ?

src/app/index.scss
```diff
@@ -9,17 +9,104 @@ $fa-font-path: "../../bower_components/fontawesome/fonts";
...
+    margin: 10px;
```
kanitw: I guess we should remove `margin: 10px` as it doesn’t have any effect on the screen from my test.   
(`margin-top|bottom` do not matter here and `.command` already defines `margin-left`, while `margin-right` is overridden on line 54).

Please let me know if I overlook anything.

vlandham: i've updated to remove this.

kanitw:
@vlandham  FYI, here is an example screen that bugs me regarding visual hierarchy for the right pane.  
That said, given we have limited time, let’s merge this so you can focus on welcome screen stuff.  

data_voyager

kanitw:
After all, these changes are already a huge improvement from the existing UIs!  Thank so much :)

vlandham:
i hear you about the hierarchy. I would suggest that they at least move forward with distinguishing the two types of visualizations presented, and perhaps more importantly help provide some context to what the user is looking at when they get started with the tool. 

Thanks again for all the discussion and thoughtfulness in working through all this.

kanitw:
>  I would suggest that they at least move forward with distinguishing the two types of visualizations presented, and perhaps more importantly help provide some context to what the user is looking at when they get started with the tool.

Yep. :)

PR #196, opened 2015-10-14
Re-render the visualization list when dataset changes

### LINKED ISSUE #176: Need to move cursor to trigger rendering
So when you select a field and not move the cursor, nothing happens. The new gallery of visualizations loads as soon as you move the cursor.

kadamwhite:
This closes #176, where the visualization list would not automatically re-render until you mouseover a pill. The reason this was happening was that the update method is debounced, and debounced methods execute outside of Angular's regular digest cycle. As such, while the Visrec object was updated, the change would not take effect until a mouseover or some other angular event handler fires, triggering a new $digest.

We could solve this problem by manually `scope.$digest`ing, but this is the use-case `$apply` was intended for: wrapping the update method in `$apply` is the most concise solution to the problem.

toggle

### DIFF
--- a/src/components/vislist/vislist.js
+++ b/src/components/vislist/vislist.js
@@ -54,7 +54,11 @@ angular.module('voyager')
           Visrec.update.projections(fieldList);
         }
 
-        var dUpdateFields = _.debounce(updateFields, 200, {maxWait: 1500});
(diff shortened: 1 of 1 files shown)


kanitw:
Nice fix!

PR #199, opened 2015-10-15
Implement Welcome Modal in Voyager

kadamwhite:
See vega/vega-lite-ui#111


src/app/main/main.html
```diff
@@ -46,6 +46,9 @@
       </div>
     </div>
     <div class="pane vis-list-pane">
+      <welcome-modal>
+        <h1>Welcome to Voyager</h1>
+      </welcome-modal>
```
kadamwhite: This could be abstracted into its own `<voyager-welcome-modal>` template, but it was easy to make a `<welcome-modal>` directive that encapsulates the localStorage behavior so I went ahead and did so

vlandham:
updated with a first pass at a welcome modal for voyager. still need the tutorial portion figured out. thoughts?

screen shot 2015-10-15 at 12 25 54 pm

kadamwhite:
The button layout broke on medium-size windows, so I have increased
the available horizontal space below a certain screen size.

image

image

image

kadamwhite:
Taking this over to rebase

kadamwhite:
`welcomeModal` and `voyagerWelcomeModal` have been combined: I am now considering this complete

domoritz:
Right now there is no way to reopen the welcome dialog once it's been closed. Should the welcome dialog be the same as the about dialog?

kadamwhite:
@domoritz That could be done—`Modals.open('welcome-modal')` will re-open the modal, so it could be re-opened to do double-duty. However, I don't think it would be inappropriate to leave this as a dedicated first-run UI and to therefore make a second About modal, elsewhere in the app; that way the purposes of the two could be customized further.

Either way I think this is in a place where, if you're comfortable with them, both this and #vega/vega-lite-ui#111 could be merged and we could iterate from there!

PR #201, opened 2015-10-15
Minor changes to `jim/update-ui` 1

kanitw:
- Center controls vertically 
- Edit copy text for univariate summary to be consistent with when a variable is selected
- Rename `.desc`, `.explain` to `.description` and `.explanation` for consistency
- add `.no-right-margin` to `.data-card` (this was my old bug)
- move `ul.menu` to be a part of `#footer` and use default 12px font-size
- "Current Data:" => "Data" 
- resize both logo — the text for voyager is now smaller and thus have better ratio between the icon and the text
- do not make logo a `.pane` remove the unnecessary `.top-pane`

There are some more minor changes coming in another PR.

data_voyager


vlandham:
`idl-logo` is different in voyager cause it was moved around
h2 is different in voyager because i took off the all caps in the css.

kanitw:
Cool, I think the h2 and idl-logo can be resolved when we migrate polestar’s style.  
Feel free to merge this to `jim/update-ui` and then merge `jim/update-ui` to master  :)

PR #202, opened 2015-10-15
Add initial in-app walkthrough using Angular Tour

vlandham:
Here is the start of an in app tutorial! Please suggest alternative wording and additional areas you believe should be part of the tutorial. Currently there are 4 tool tips. tooltip scss based on included css. can certainly be tweaked more. It should perhaps be put somewhere in vlui eventually?

screen shot 2015-10-15 at 4 20 08 pm

### DIFF
--- a/src/app/index.scss
+++ b/src/app/index.scss
@@ -1,7 +1,4 @@
-// FIXME should use normalize.scss version instead
-// but I got some errors so I don't want to spend too much time here now.
 @import "../assets/normalize.scss";
-@import "../assets/angular-tour.scss";
 
 // injector
 // endinjector
--- a/src/app/index.scss
+++ b/src/app/index.scss
@@ -1,6 +1,7 @@
 // FIXME should use normalize.scss version instead
 // but I got some errors so I don't want to spend too much time here now.
 @import "../assets/normalize.scss";
+@import "../assets/angular-tour.scss";
 
 // injector
 // endinjector
--- a/src/app/main/main.html
+++ b/src/app/main/main.html
@@ -36,7 +36,7 @@
         <div id="footer">
           <ul class="menu">
             <li>
-              <a ng-click="">About</a>
+              <a ng-click="startTour()">Show Tour</a>
             </li>
             <li ng-if="consts.debug">
               <a class="debug" ng-click="toggleDevPanel();">Debug</a>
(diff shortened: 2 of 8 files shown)


src/app/index.scss
```diff
@@ -1,6 +1,7 @@
 // FIXME should use normalize.scss version instead
 // but I got some errors so I don't want to spend too much time here now.
 @import "../assets/normalize.scss";
```
kadamwhite: Not related to @vlandham's work, but it seems like we *do* use `normalize.scss` here—is this comment out of date?

src/app/main/main.html
```diff
@@ -66,4 +72,42 @@
     </bookmark-list>
   </modal>
   <dataset-modal></dataset-modal>
+  <tour step="currentStep">
+    <virtual-step
```
kadamwhite: This may be overkill, but it might make maintenance easier to require this markup in via an [ng-include](https://docs.angularjs.org/api/ng/directive/ngInclude)

src/assets/angular-tour.scss
```diff
@@ -0,0 +1,138 @@
...
+  transition: background-color 300ms ease-out;
+  cursor: pointer;
+  font-family: inherit;
+  font-weight: 700;
+  line-height: 1;
+  position: relative;
+  text-decoration: none;
+  text-align: center;
+  display: inline-block;
+  font-size: .8em;
```
kadamwhite: This is a little small, it comes across to me as visually cramped:
image

I might remove this rule, the button isn't so much bigger without the smaller font that I think it outweighs the legibility compromise

kadamwhite:
I recorded the jQuery version resolution in the bower config, which should fix the broken build here.

This looks great, and the tooltips are working for me! This will be easy to wire up to #199 as well, which should make on-boarding users much nicer. I defer to @kanitw, @domoritz & @jheer on the wording, but from a code standpoint, :+1: !

src/app/index.scss
kanitw: Looks like it is.

src/assets/angular-tour.scss
kanitw: We should also avoid introducing too many different font-size in the app for visual consistency.   I have to admit that I haven't done very good job on this regard during the pre-user-study crunch, but we should avoid making it worse :)

kanitw: And +1 on the button being too small

kanitw:
1) I checked out your branch, but can only see the first tooltip. When I click next, the first tooltip just disappear but the second doesn't show.  Not sure what I did wrong.  

2) Regarding copy text:
For the first tooltip, we actually should point out just that the left side shows schema, with data type icon and  has tooltip about the field's information. 

However, instead of saying:

>  You can tell it what fields to start visualizing by selecting them here. Data Voyager will use these fields as a starting point for more recommendations.

The prompt suggests people to start selecting variable right away.  This might lead users to early fixation with a few variables -- which is opposite to Voyager's design goal to encourage broader exploration. 

Maybe we should actually highlight the right side and suggests people to look at univariate summary first to familiar themselves with dataset first.  Or at least just say that they can "select variable to steer the recommendation", but avoid hinting that they should select to start.

I will punt on commenting on other tooltip and will do it after I get them to work, so I can comment them in context (which is easier than digging the code as guess where do they point at).  

You might want to check out [Voyager video](https://vimeo.com/135417594) to see how we demo the app.   It won't be 1:1 mapping to what we want to show here as we actually interact with the app in the video, while here is a set of wizard that let users click next.  

3) Given that we're running out of time, we can take over on copy editing if you don't get to finish it. 
Having implementation in place is already __AWESOME__!

src/app/main/main.html
vlandham: I'm not opposed to it. suggestions for where to put the `.html` ? I could create a new sub-directory under `components` to store this and the scss? I could just add it to assets?

kadamwhite: > I could create a new sub-directory under `components` to store this and the scss

I think that would be my instinct. But again, this may not be a priority :)

vlandham:
thanks @kadamwhite and @kanitw for all the feedback! i've updated with the PR to move the tour out into its own component. I've also removed a lot of the font stuff from the scss. @kanitw you are right about the text. I've tried to re-word to improve it - but we can continue to iterate. 

very strange that you are only able to see the first tooltip. I will start trying it out on other browsers. 

screen shot 2015-10-16 at 9 28 34 am

kadamwhite:
@kanitw I just force-pushed this, so I'd
```bash
git fetch origin
git branch -D jim/add-tutorial
git checkout jim/add-tutorial
```
to get a new reference to it before you test again.

kanitw:
@kadamwhite  thanks for the heads up

Issue #205, opened 2015-10-16
Consider changing footer menu to be inline-block with some delimiter such as  · 


kanitw:
See Facebook as an inspiration

![facebook](https://cloud.githubusercontent.com/assets/111269/10531039/080f6346-7362-11e5-800f-edab953e11dc.png)

kanitw:
Done in #208

PR #208, opened 2015-10-16
Try Implementing inline-block footer links as described in #205

kadamwhite:
image

Note that this does work better for many links when there is sufficient room: if we add in the tour link, and debug mode is enabled (otherwise those other two don't display), I don't know any way to avoid that hanging bullet on the first line:

image

Facebook's solution (from #205) has the same problem, and it wouldn't happen in debug mode, so this may not be an issue at all.

Alternate option for the separator would be the smaller dot `\000b7` instead of the bullet &bull;, which might actually be what FB uses:

image

### DIFF
--- a/src/app/index.scss
+++ b/src/app/index.scss
@@ -135,10 +135,10 @@ h2 {
       padding: 0;
       display: inline-block;
 
-      // items Separate with a white bullet
+      // Separate items with a dot
       &:after {
         color: #eee;
-        content: '\2022';
+        content: '\000b7';
         padding: 0 3px;
       }
       &:last-of-type {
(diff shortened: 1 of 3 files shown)


kanitw:
LGTM.   Thanks for changing this!  :) 

Minor: I like the smaller dot better than bullet, but no strong preference.  Feel free to pick one and merge.

kadamwhite:
@kanitw Great! I switched to the dot (I liked it slightly more as well) and fixed the debug button, which I'd broken by accident in my first commit.

This should probably be made into another issue, but cc @vlandham in case this falls under the purview of #199 : the "About" button @vlandham added has no function, and does not appear to ever have had any. What's the plan there?

kanitw:
> the "About" button

We will add about dialog, showing brief description of the project and credit everyone who involves in this project.  (Bocoup, IDL, Tableau Research)

PR #209, opened 2015-10-16
Workaround for tooltip opacity issue (fixes #200)

### LINKED ISSUE #200: Many “visible” but transparent tooltips block “hover” on command buttons
Right now we can’t hover/click the command button for the top-left visualization in voyager 

This is recently introduced as we do not have the same problem in the [current online demo](http://vega.github.io/voyager)

This is happening both in master and the `jim/update-ui` branch

pasted_image_10_15_15__11_38_am

pasted_image_10_15_15__12_07_pm

The root cause is probably that now those transparent `._720kb-tooltip` has `visibility: visible;` while they had `visibility: hidden` in the online demo.  

@domoritz @kadamwhite @vlandham  — any idea what’s causing this?

kadamwhite:
The CSS intended to hide a tooltip when it is not open does not work due to a CSS selector precedence issue in the angular-tooltip code. Using `!important` is generally not best-practice, but it does let us temporarily work around the problem described in issue #200 and 720kb/angular-tooltips#95

Once this issue is resolved in the angular-tooltips library this code can be deprecated and removed.

### DIFF
--- a/src/app/index.scss
+++ b/src/app/index.scss
@@ -150,3 +150,8 @@ h2 {
     }
   }
 }
+
+// Workaround for https://github.com/720kb/angular-tooltips/issues/95
+._720kb-tooltip:not(._720kb-tooltip-open) {
+  visibility: hidden !important;
+}


kanitw:
Just check and looks like this is a great temp solution.  Thanks for fixing this!

PR #222, opened 2015-10-23
Show welcome modal on click on “about”

domoritz:


kanitw:
Ideally welcome should be different from about but I think this is better than nothing for now.

PR #317, opened 2017-05-16
Add redux-logger to development build.

vlandham:
Imports redux logging for development builds:

screen shot 2017-05-16 at 10 17 22 am

I didn't see a production or deploy script for the tool yet - so I didn't test to ensure it would be compiled out for production builds. The:

```
if (process.env.NODE_ENV !== 'production') {
```

is what we have used in the past. If there is a different way that a production build is indicated - let me know, and i will change this.


kanitw:
> I didn't see a production or deploy script for the tool yet 

Yeah, I haven't setup production config at all.

vlandham:
Sorry - previous build issues were because I brought in `@types/node` - but node's `setTimeout` has a different signature than client JS. 

In light of this - and because we don't need all node types, I've just created a `process` type in the store. we can move this out of store if this is needed elsewhere.

Issue #313, opened 2017-05-15
Add APIs to load dataset from outside 


kanitw:
## Current Behavior

To give some context of what the app is currently doing: 

1) When the application start, we load the [`DEFAULT_DATASET`](https://github.com/vega/voyager/blob/vy2/src/models/dataset.ts#L16).  

2) We allow users to change the dataset using the [DatasetSelector](https://github.com/vega/voyager/blob/vy2/src/components/data-pane/dataset-selector.tsx), which in turn calls [`datasetUrlLoad`](https://github.com/vega/voyager/blob/vy2/src/actions/dataset.ts#L28), which is a `ThunkAction`.

This `datasetUrlLoad` calls CompassQL's buildSchema and dispatches the `DATASET_URL_RECEIVE` to update the `dataset` in the model. 

Note: 
- Note that the default dataset and the Dataset Selector UI are not ideal behavior -- they are just placeholder for initial development. 
- I only set up React's asynchronous using redux-thunk because it's included in the main redux tutorial, but if you have a good reason why we should choose other asynchronous action pattern, we can easily swap to other as we're still in an early stage. 

## What to do

- [x] Design an API (probably in [`app.tsx`](https://github.com/vega/voyager/blob/vy2/src/components/app.tsx)) to have a command to load dataset from outside -- either as a URL, or loading the data directly. 

- [ ] We probably have to refactor the `dataset` in the model somehow to make it more flexible and better support server-side (https://github.com/vega/voyager/issues/314).

- [ ] (Advance) How should our application be synchronized with external data if it can be changed -- e.g., with Jupyter Lab)

tafsiri:
@kanitw and @vlandham i've made a few issues that relate to this but are more granular (listed above). Feel free to take a look and comment. I think it would be nice to close this issue out in favor of more granular ones. The second bullet above seems covered by #314. The third one might also have solution, but we'd probably have to evaluate that in the light of an embedding.

PR #319, opened 2017-05-16
Add wrapper for voyager app and allow programmatic setting of data source

tafsiri:
~~~_Do not merge (probably)_~~~

New build target has been added that aims to build the voyager app as a  consumable library.

The new entry point in voyager.tsx wraps the react app, and allows instantiating it in another app without direct use of react. updateData and updateConfig are stubs for now.

To use a new task has been added `yarn run build-lib`

This will create a lib folder with a bundle.js and style.css. These can be used in another js app (bundle should be UMD) but I've only tried it with require.js so far. The resulting build is currently pretty big but can be optimized. 

Next step for me is investing what updateData and updateConfig would do in voyager.tsx

Posting this to keep in sync with @vlandham and @kanitw. Is this in line with what we were discussing?


tafsiri:
This branch will be ready for review once #322 is merged into it. 

I'm removing the WIP tag as it demonstrates getting data into the app programmatically. It pretty much #313 but i'd like to split that issue out into more granular ones that would enhance this functionality.

tafsiri:
I also think this will pass CI once #322 merges into this branch.

tafsiri:
This is good to go for review. I started down the path of adding a test for this, but running into some issues with jest and the various webpack loaders, so that may be a while coming. If this is good to merge i can follow up later on that.

tafsiri:
Note that this uses a few `any` types in places.

src/lib-voyager.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * The Voyager class encapsulates the voyager application and allows for easy
+ * instantiation and interaction from non-react projects.
+ */
+class Voyager {
+  private container: HTMLElement;
+  private config: VoyagerConfig;
+  private data: VoyagerData;
+  private store: any; // TODO how to get the type/inteface for this here?
```
kanitw: I hover over configureStore() in VSCode to see the return type and try to import things accordingly.

src/components/app.tsx
```diff
@@ -5,13 +5,43 @@ import { DragDropContext } from 'react-dnd';
...
+
+interface VoyagerAppProps extends ActionHandler<DatasetAsyncAction> {
+  config: Object;
+  data: Data;
+  dispatch: any;
+}
+
+class AppBase extends React.PureComponent<VoyagerAppProps, {}> {
+
+  constructor(props: any) {
```
kanitw: @tafsiri  I tried changing `any` to `VoyagerAppProps` and this throws a reasonable error.  

Basically `VoyagerAppProps extends ActionHandler<DatasetAsyncAction>` but your call of App in both index and lib-voyager does not have the right property required by `ActionHandler<DatasetAsyncAction>`.   

I haven't thought much about whether you need to extend `ActionHandler<DatasetAsyncAction>`here, but I have to take back what I say yesterday.  I think the error from any here is a different problem from what I see error.  So I'd try to remove any and add the right type and try to correct things to match.

kanitw: If you want me to further look on this, let me know!

src/components/app.tsx
```diff
@@ -57,22 +57,20 @@ class AppBase extends React.PureComponent<any, {}> {
     );
   }
 
-  private update(props: any) {
+  private update(props: VoyagerAppProps) {
     const { data } = props;
-    if (VegaLiteData.isUrlData) {
+    if (isUrlData(data)) {
```
kanitw: @tafsiri FYI, type guard are methods so you have to call them too.  

Also, i found basarat.gitbooks.io/typescript/ useful when I started working on TS. :)

src/components/app.tsx
```diff
@@ -5,13 +5,43 @@ import { DragDropContext } from 'react-dnd';
...
+  datasetUrlLoad,
+} from '../actions';
+
+
+export type VoyagerConfig = Object;
+export type VoyagerData = InlineData | UrlData;
+
+interface VoyagerAppProps extends ActionHandler<DatasetAsyncAction> {
+  config: Object;
+  data: Data;
```
kanitw: @tafsiri  I wonder if we could make this `data` named `fixedData`.
Basically, if there is a fixed dataset, we just don't show data selector and just show this fixedData.  Otherwise, show a data selector like normal.

tafsiri: i think it would be better to set that explicitly in the config. (whether to show data selector or not)

tafsiri: i think that better separates out the concerns of whether you want an initial dataset to be something in particular vs ability to switch between things. I don't think the UI part for the selector should depend on the currently set data.

tafsiri:
Ok, I've added more type information into the app component, as well as some general cleanup.

However there are some things that I don't think are fixable/typable in the file. I'll leave some comments inline.

src/components/app.tsx
```diff
@@ -5,13 +5,41 @@ import { DragDropContext } from 'react-dnd';
...
+
+interface Props extends React.Props<AppBase> {
+  config?: VoyagerConfig;
+  data?: Data;
+  dispatch: Dispatch<StateWithHistory<Readonly<StateBase>>>;
+}
+
+class AppBase extends React.PureComponent<Props, {}> {
+
+  constructor(props: any) {
```
tafsiri: This needs to be `any`. If you change it to `Props` it will not compile. After much much searching and experimentation, I believe it is because of the HTML5Backend decorator. See http://stackoverflow.com/questions/40269415/typescript-2-0-with-react-dnd-gives-error-jsx-element-attributes-may-not-be-a-u for more details. 

I also think this is what prevents this being compatible with connect. So I am passing dispatch in directly.

Though do feel free to take a look in case i've overlooked something.

tafsiri: Also I've left the props optional for now as the app does not depend on them being passed in. As the config and other elements get fleshed out we can make them required and pass them in explicitly in all cases.

tafsiri:
Ready for re-review @kanitw, I've probably gone as far as I can with the type info for props.

tafsiri:
cc @vlandham

src/components/app.tsx
kanitw: Ah I think this `HTML5Backend` is the same issue that I had earlier

src/components/app.tsx
```diff
@@ -5,13 +5,41 @@ import { DragDropContext } from 'react-dnd';
...
+    super(props);
+  }
+
+  public componentWillUpdate(nextProps: Props) {
+    this.update(nextProps);
```
kanitw: Wondering if we have to compare with the previous state to avoid unnecessary load request?

src/lib-voyager.tsx
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Update the configuration of the voyager application.
+   *
+   * @param {VoyagerConfig} config
+   *
+   * @memberof Voyager
+   */
+  public updateConfig(config: VoyagerConfig) {
+    this.config = config;
+    this.render(this.data, config);
```
kanitw: I wonder if changing config should always re-render the whole app?  
(I guess this seems okay for now, but just want to ask to verify)

src/components/app.tsx
tafsiri: PureComponent provides an implementation that shallow compares with previous props. At this point I think that should be sufficient. We can override that implementation later down the road if necessary, though i don't think we will need to given how fundamental data and config are to what is rendered.

src/lib-voyager.tsx
tafsiri: Yes I think it should re-render the whole app, that way anything that depends on these props (or things derived from them) in the store will update accordingly. I might be missing a nuance to your question so let me know. However maybe i should add that calling ReactDOM.render in the code below is just like rendering a component further down the state tree, react will do dom-diffing, and components will run their regular lifecycle methods etc, it wont 'force' a fresh render so to speak (e.g. removing the current dom elements completely and starting from scratch).

Issue #250, opened 2017-02-09
Setup Webpack-Dev-Server with HMR


kanitw:
Currently webpack-dev-server somehow does not update the build properly.

(For now, I'm using `npm run watch` and use `python -m SimpleHTTPServer <port>` to serve the app.)

## TODO

- [ ] Make webpack-dev-server at least compile properly.  This should allow us to modify a file and refresh the browser to get latest build.) 
- [ ] Support Hot Module Replacement (Update Content without refresh)
  - [ ] CSS
  - [ ] React Components

## Potentially Useful Resources

 http://gaearon.github.io/react-hot-loader/
https://medium.com/@rajaraodv/webpacks-hmr-react-hot-loader-the-missing-manual-232336dc0d96#.aytuhxvht
https://webpack.js.org/guides/hmr-react/
(Not sure if it is already supported by Redux though.)

kanitw:
cc: @felixcodes -- if you're looking for a challenging and extremely helpful issue, this is one!

(This will obviously speed up our development a lot in long run.)

saulshanabrook:
>Currently webpack-dev-server somehow does not update the build properly.

I just tried starting the dev server with `npm run start` and it seems to be updating when I change a file:

![](https://fat.gfycat.com/ImmediateHeftyGrayfox.gif)

Could you clarify what was broken? I am happy to start working on HMR if this part is done already.

tafsiri:
This is also working on my end using both yarn start and npm start. If @kanitw you have a moment to confirm we can probably close this.

kanitw:
Yeah it's working.  I'll keep this open and rename the issue as it's still nice to have HMR.

PR #320, opened 2017-05-16
Add production build script

vlandham:
This is a start to integrating the latest build processes from [create-react-app](https://github.com/facebookincubator/create-react-app). 

Not sure it is worth all the additional development packages, but it does provide auto-reloading on code edits for the `watch` process and provides a production build that (mostly) works.

Could be improved with:

* adding TS Lint errors to the development output
* separating `bundle` and `vendor` js bundles again
* copying over vega-datasets on the production build so that it can still load data.


vlandham:
thoughts @tafsiri

vlandham:
well, the additional packages might not be compatible with the travis system anyways:

> error react-dev-utils@1.0.0-alpha.e280254d: The engine "node" is incompatible with this module.

kanitw:
I don't like copying *all* scripts from create-react-app as it's adding so many new dependencies and make the build way harder to maintain especially create-react-app is created for JS-based application, not TS-based.  For example, I don't think we really need to add `babel` as we use typescript compiler.

Maybe better just to copy the parts we need?

vlandham:
I agree we should just copy what we need. I'll revise with removal of the babel stuff - I was just trying to see if it would work without a lot of modification. 

Jim

tafsiri:
@vlandham sounds good. I'm thinking the necessary bits could be added to the existing webpack config. What are the main things you see need to be done for the prod build. As I see it: 

1: is setting some variables in the code so that the code can take an appropriate path when needed (e.g. setting up the logger or not). I wonder if using something like the [DefinePlugin](https://webpack.js.org/plugins/define-plugin/) or [EnvironmentPlugin](https://webpack.js.org/plugins/environment-plugin/) before the ts-loader kicks in would work?

2: is doing minification and stuff. I wonder if this can be done on the bundle created by tsc by something like uglify or closure compiler. 

Are there other things you were thinking of?

vlandham:
I would say we want:

for dev build:

* enable hot reloading for js and scss changes
* have only one command to run to start watch and serve processes
* if possible - display compile time errors in html 
* setup env. variables.

For production build:

* minification / uglification
* hashing build files - to prevent caching issues. 
* some mechanism to bring in static files (like the example data files) into the build
* something that can easily be extended for deploys
* setup env. variables.

tafsiri:
@kanitw relating to Jim's point about hot code reloading, is #250 out of date. I also have been using webpack-dev-server through yarn start and its been working well. Maybe contributing.md needs an update?

kanitw:
@tafsiri  I haven't got a chance to test in the past few weeks yet-- it's weird that somehow it suddenly works, but if so, feel free to update #250.

vlandham:
This now provides a working production build of voyager. The dev build has been reverted to use webpack-dev-server. 

The production build minifies all the code and also copies over the voyager-datasets data into the `dist` directory. 

One issue I haven't quite figured out is how to use the `HtmlWebpackPlugin` for both builds. It works well for production build - injecting the generated css / js files into the html template. But doesn't work for the dev build.

For now, there is a separate `index.ejs` used for the template file of the production build.

vlandham:
closes #321

index.ejs
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
kanitw: I guess this file is intended to follow `http://www.embeddedjs.com/` thus name `.ejs`?

vlandham: By default the [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin#writing-your-own-templates) uses the [lodash template loader](https://github.com/okonet/ejs-loader). This doesn't really impact us much - and we can change the naming scheme. Ultimately i'd like to combine index.ejs and index.html - but need to do a bit more research as to why the plugin wasn't working in dev.

PR #322, opened 2017-05-17
313 set data store prop

tafsiri:
Allow data to be set via prop. Note that this PR is going into the embed wrapper branch.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -10,35 +10,41 @@ import {EncodingPane} from './encoding-pane';
 import {Header} from './header';
 import {ViewPane} from './view-pane';
 
-import {connect} from 'react-redux';
-import {State} from '../models';
 import {
-  createDispatchHandler,
-  datasetUrlLoad,
   ActionHandler,
   DatasetAsyncAction,
+  datasetUrlLoad,
 } from '../actions';
 
-export type VoyagerData = string | Object[];
+
+
+interface DataURL {
+  name: string;
+  url: string;
+};
+export type VoyagerData = DataURL | Object[];
 export type VoyagerConfig = Object;
 
 interface VoyagerAppProps extends ActionHandler<DatasetAsyncAction> {
   config: Object;
   data: VoyagerData;
+  dispatch: any;
 }
 
-class AppBase extends React.PureComponent<ActionHandler, {}> {
+class AppBase extends React.PureComponent<any, {}> {
+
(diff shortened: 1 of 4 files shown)


tafsiri:
ping @vlandham

tafsiri:
essentially i'd like to get this code into 318-embed-wrapper and then review that as a whole

vlandham:
looks good to me.

src/components/app.tsx
```diff
@@ -10,8 +10,43 @@ import {EncodingPane} from './encoding-pane';
...
+  datasetUrlLoad,
+} from '../actions';
+
+
+
+interface DataURL {
+  name: string;
+  url: string;
+};
+export type VoyagerData = DataURL | Object[];
```
kanitw: Also, wonder if there is any reason to diverge from `Data` object in Vega-Lite yet.  

If not, we could use consistent structure and use `isUrlData`, `isInlineData` type guards provided in `vega-lite/src/data`

tafsiri: Oh interesting. Didn't know about the vega-lite data object. Since doing this i've been thinking of what a consistent format could be and started talking about that in #324, basically `{ "type": "dataset", name: "Foo", "values": [] }`. But i'll look at that interface/type in vega-lite and see if that can suffice.

src/components/app.tsx
```diff
@@ -10,8 +10,43 @@ import {EncodingPane} from './encoding-pane';
...
+export type VoyagerConfig = Object;
```
tafsiri: Sure. At the moment its just empty since we don't know whats going into it (though stuff like disabling the dataset selector or not would be there)

tafsiri: looks like tslint disallows empty interfaces. I'll keep in mind changing this to a minimal interface when i actually start working on the config.

src/components/app.tsx
tafsiri: @kanitw this has been done. with one caveat. it seems that UrlData or InlineData do not have names. For now i've just hard coded "custom data" as a name. I can get your thoughts in another thread of why NamedData is separate and whether the other two data types could have optional names. (Or whether i should just extend one with the other—though i feel like that would be a new custom type, which i am trying to avoid if possible).

tafsiri:
Thanks for the review @kanitw I'm going to merge this into my other branch and try clean it all up for final review.

Issue #314, opened 2017-05-15
Add Server Mode


kanitw:
Currently, all query processing happens in the client.

- When a new dataset is loaded, we call the [`buildSchema` method](https://github.com/vega/voyager/blob/vy2/src/actions/dataset.ts#L42) to create a schema from a dataset and calculate basic statistics. 

- When the shelf updates, we call CompassQL's [`recommend` method directly in the selector](https://github.com/vega/voyager/blob/vy2/src/selectors/index.ts#L24).  (Once I implement related views, we will also make multiple calls to the CompassQL to generate other recommendations as well.) 

It would be great to add a server (probably express.js) to allow this computation to happen separately on a server. 

Note that we should still support running in a client-only mode.

domoritz:
Can you elaborate what this is?

PR #328, opened 2017-05-18
327 Move compassql recommend results to data store

### LINKED ISSUE #327: move recommend into data store
As part of #314 , the `recommend()` functionality of compassql needs to be able to be called sync and async.

vlandham:
closes #327 . As a prereq to #314, this moves the call to compassql's `recommend()`to the data store, and allows for it to be called synchronously or asynchronously. 

Currently the recommendations are stored in the datastore in a sub-tree called `compass`. If that name is ambiguous, we can change it. I plan on moving other compassql function results into this data store as well (basically anything that will be returned by the server can go in here). 

The dispatch to get new recommendations is performed whenever the dataset or the shelf changes. Right now these dispatches happen in their respective components. We also discussed automatically dispatching at the redux store level - but that is generally frowned upon.


vlandham:
looks like i messed up the rebase - sorry for the additional commits.

src/models/index.ts
```diff
@@ -1,17 +1,20 @@
...
 export * from './shelf';
+export * from './compass';
 
 /**
  * Application state.
  */
 export interface StateBase {
   dataset: Dataset;
   shelf: Shelf;
+  compass: Compass;
```
kanitw: Let's make this `results: ResultIndex` as it is essentially a index of results from CompassQL Query

src/models/compass.ts
```diff
@@ -0,0 +1,12 @@
+import {SpecQueryModelGroup} from 'compassql/build/src/model';
+
+export interface Compass {
+  isLoading: boolean;
+
+  recommends: SpecQueryModelGroup | null;
```
kanitw: Let's name this `modelGroup`.

src/actions/compass.ts
```diff
@@ -0,0 +1,45 @@
...
+import {Dispatch} from 'redux';
+import {ThunkAction} from 'redux-thunk';
+import {ReduxAction} from './redux-action';
+
+import {fetchCompassRecommends} from '../api/api';
+import {State} from '../models/index';
+import {getQuery, getSchema} from '../selectors';
+import {Action} from './index';
+
+export type CompassAction = CompassRecommendsRequest | CompassRecommendsRecieve;
```
kanitw: I think it is better to use `Query` prefix than `Compass` as we are querying visualizations from CompassQL (In any case, it's CompassQL, not the old Compass anyway).

src/actions/compass.ts
```diff
@@ -0,0 +1,45 @@
...
+export type CompassAsyncAction = CompassRecomendationsLoad;
+
+export const COMPASS_RECOMMENDS_REQUEST = 'COMPASS_RECOMMENDS_REQUEST';
+export type CompassRecommendsRequest = ReduxAction<typeof COMPASS_RECOMMENDS_REQUEST, {}>;
+
+export const COMPASS_RECOMMENDS_RECEIVE = 'COMPASS_RECOMMENDS_RECEIVE';
+export type CompassRecommendsRecieve = ReduxAction<typeof COMPASS_RECOMMENDS_RECEIVE, {
```
kanitw: Recieve => Receive

QueryResultReceive / QueryRecommendationReceive / QueryReceive?

The latter seems shortest -- so I like that the most.

src/api/api.ts
```diff
@@ -0,0 +1,11 @@
+
+import {Query} from 'compassql/build/src/query/query';
+import {recommend} from 'compassql/build/src/recommend';
+import {Schema} from 'compassql/build/src/schema';
+
+export function fetchCompassRecommends(query: Query, schema: Schema) {
```
kanitw: fetchCompassQLResults / fetchCompassQLRecommendations ? 

(The former is shorter)

src/actions/compass.ts
```diff
@@ -0,0 +1,45 @@
...
+  recommends: SpecQueryModelGroup
+}>;
+
+export type CompassRecomendationsLoad = ThunkAction<void , State, undefined>;
+export function compassRecomendationsLoad(query?: Query, schema?: Schema): CompassRecomendationsLoad {
```
kanitw: I know I name a similar method for datasetUrl `datasetUrlLoad`  -- but I start wondering if we should come up with a better name for methods that does request + receive.

("load" doesn't really suggest request + receive.)

src/components/encoding-pane/index.tsx
```diff
@@ -6,26 +6,38 @@ import {Channel} from 'vega-lite/build/src/channel';
...
     // Bind - https://facebook.github.io/react/docs/handling-events.html
     this.onClear = this.onClear.bind(this);
   }
+
+  public componentDidMount() {
+    this.props.handleAction(compassRecomendationsLoad());
+  }
+
+  public componentDidUpdate() {
+    this.props.handleAction(compassRecomendationsLoad());
```
kanitw: This will cause the asyncQueryRequest (compassRecomendationsLoad)  when `specPreview` changes as well.

src/selectors/index.ts
```diff
@@ -19,9 +18,9 @@ export const getQuery = createSelector(
 );
 
 export const getMainResult = createSelector(
-  getQuery, getSchema, getData,
-  (query: Query, schema: Schema): SpecQueryModelGroup => {
-    return recommend(query, schema).result;
+  getRecommends,
```
kanitw: If we rename `getRecommends` to `getMainResult`, I think we don't really need this composite selector anymore.

src/actions/compass.ts
kanitw: Suppose we name the subactions `QueryReceive/Request`  -- maybe this method could be `asyncQueryRequest`?

src/components/data-pane/dataset-selector.tsx
```diff
@@ -29,6 +29,10 @@ export class DatasetSelector extends React.PureComponent<DatasetSelectorProps, {
     this.onDatasetChange = this.onDatasetChange.bind(this);
   }
 
+  public componentDidUpdate() {
```
kanitw: Not sure why we need to trigger this action here?  
I'm afraid we are going to trigger the query more often than we need to.

src/models/compass.ts
```diff
@@ -0,0 +1,12 @@
+import {SpecQueryModelGroup} from 'compassql/build/src/model';
+
+export interface Compass {
```
kanitw: Let's call this `Result` and make a 

```
export interface ResultIndex {
   /** This is the result of the query from the shelf */ 
   main: Result;  

  // TODO: Add other results for other related views in the near future.)
}
```

src/components/encoding-pane/index.tsx
kanitw: I wonder what is the standard pattern for making asynchronous request based on changes in another part of the state in the store. 

Basically, the `asyncQueryRequest` should happen when the `state.shelf.spec` changes.  
(Should that instead be in the reducer? -- I'm not sure -- just asking.)

src/models/compass.ts
vlandham: I was just thinking we might want to store more then just the results here - but also the other functions of compassql that could come from the server.

src/actions/compass.ts
vlandham: ok

src/components/encoding-pane/index.tsx
vlandham: we can research more. this is the simplest mechanism - but yes could result in unnecessary renders. I was asking the same question when implementing this. The most common / best would be to keep it all in the action creators - but since this request to get recommendations depends so much on the different pieces of the `shelf` - it seemed a waste to try to hook all those action creators to additionally call the update to recommend()

Putting it in the reducer would mean the reducer has side effects (which is frowned upon), but a similar question (https://github.com/reactjs/reselect/issues/142) points to researching [Redux Loop](https://github.com/redux-loop/redux-loop) a bit to see if there is a sane way to do it in the reducer.

An easy alternative would be to just add a bit of checking to this `componentDidUpdate` to only call if the shelf changes.

src/selectors/index.ts
vlandham: i agree. i was just keeping it out of the root `getRecommends` in case we need more values for the main panel. (also it was nice for debugging).

src/components/data-pane/dataset-selector.tsx
vlandham: I can check - but this should trigger when the data is changed.

src/models/compass.ts
kanitw: I see.  Can you be more specific about what you are thinking about?  

(In any case, I feel like Compass is a vague "catch all" name.)

src/components/encoding-pane/index.tsx
kanitw: > An easy alternative would be to just add a bit of checking to this componentDidUpdate to only call if the shelf changes.

This seems like a reasonable alternative for now so we won't get blocked. 
(We can file an issue to revise this later?)

src/actions/compass.ts
vlandham: maybe. though to the caller of these functions - i'm not sure they should care/know if the function is async or not - thats a benefit of redux-thunk - right? that async and sync actions are interoperable from a user of the action creator?

src/actions/compass.ts
vlandham: So the data store is going to be renamed `result` 

should these actions follow that naming convention and be called `ResultAction` ? instead of `Query` ?

src/components/encoding-pane/index.tsx
vlandham: i will add this

src/selectors/index.ts
vlandham: i will use suggested naming

src/actions/compass.ts
kanitw: Good question.  I think ResultRequest / ResultReceive sound good to me.  

ResultAction sounds a bit weird, but I guess that's better than having `QueryAction = ResultRequest | ResultReceive`.

src/actions/compass.ts
vlandham: I enjoyed this article from mapbox: https://www.mapbox.com/blog/redux-for-state-management-in-large-web-apps/

which had this to say about action creator names:

> Our action creators are functions named with simple imperative verb phrases, e.g. updateCreditCard, verifyEmail, uploadDataset, dismissWelcomeModal . Often these function names do not correspond to technical plumbing behind the scenes; instead, they describe an intention, a command. The action creator might actually dispatch several actions, depending on what it must accomplish. The point is that it does what you ask it to do.

kanitw: good point.  Maybe we can just call it queryRequest / resultRequest? 
(Given that users of these APIs won't care about the "receive" part.)

vlandham: I would suggest after we get some of this merged in - we can take a holistic look at the naming conventions, and make changes accordingly

src/components/data-pane/dataset-selector.tsx
vlandham: i don't believe this currently makes un-necessary re-renders

vlandham:
@kanitw i believe this is ready for re-review.

src/actions/dataset.ts
```diff
@@ -44,6 +45,7 @@ export function datasetUrlLoad(name: string, url: string): DatasetUrlLoad {
           type: DATASET_URL_RECEIVE,
           payload: {name, url, schema}
         });
+        return { schema };
```
kanitw: I'm curious why we have to return here?

src/actions/dataset.ts
```diff
@@ -23,10 +23,11 @@ export type DatasetUrlReceive = ReduxAction<typeof DATASET_URL_RECEIVE, {
   schema: Schema
 }>;
 
+
 export type DatasetUrlLoad = ThunkAction<void , State, undefined>;
 
 export function datasetUrlLoad(name: string, url: string): DatasetUrlLoad {
-  return (dispatch: Dispatch<Action>) => {
+  return (dispatch: Dispatch<Action>, getState: Function) => {
```
kanitw: Do we use getState here?

src/actions/dataset.ts
vlandham: we no longer have to. we can remove. it was for chaining promises that isn't currently used.

src/actions/dataset.ts
vlandham: nope! not any more. my bad.

PR #331, opened 2017-05-19
323 load live data

tafsiri:
Add a new action to load InlineData into voyager programmatically. 

cc @kanitw @vlandham


tafsiri:
also, while it looks like there are a lot of commits, that is because i had branched off of my prev branch before it was merged. This would all be squashed down for merging imo.

tafsiri:
and the github button now says that... how convenient

src/components/app.tsx
```diff
@@ -8,7 +8,7 @@ import * as SplitPane from 'react-split-pane';
 import { Dispatch } from 'redux';
 import { StateWithHistory } from 'redux-undo';
 import {Data, InlineData, isInlineData, isUrlData, UrlData} from 'vega-lite/build/src/data';
-import { datasetUrlLoad } from '../actions';
+import { datasetReceive, datasetUrlLoad } from '../actions';
```
kanitw: FYI, (a very very minor point) our codebase currently uses no space between {} in imports.  

(Maybe we can setup tslint and some autoformatter to fix this. I think we did set this in Vega-Lite, but I haven't ported the setup over.)

src/actions/dataset.ts
```diff
@@ -48,4 +49,23 @@ export function datasetUrlLoad(name: string, url: string): DatasetUrlLoad {
       }
     );
   };
-}
+};
+
+export const DATASET_RECEIVE = 'DATASET_RECEIVE';
```
kanitw: Wonder if we should name this `DATASET_INLINE_RECEIVE`?

tafsiri: We could, though i've been wondering if we should unify the `DATASET_URL_RECEIVE` and this? It has not been clear to me what the app does with either the URL or the original data (in the case of inline data), it puts it in the store, but i was under the impression that the schema also encapsulates the data? 

If we unified the two then the action creator for urlReceive would just build a schema and dispatch this single action. 

What do you think?

tafsiri: Though for this PR i can rename it. (I wouldn't do the unification in this PR)

tafsiri:
@kanitw. PR has been updates. I had left a comment about DATASET_INLINE_RECEIVE and DATASET_URL_RECEIVE for your later consideration.

src/actions/dataset.ts
kanitw: Thanks.  These are great questions.  :) 

> It has not been clear to me what the app does with either the URL or the original data (in the case of inline data), it puts it in the store

Currently, we just stick the `data` in the rendered specifications. 
So for UrlData, Vega runtime will read from the url again to render each plot. 
For InlineData, Vega just reads the data inline.  

Eventually, we would change both cases to bind data at runtime so that we can reduce redundant transformation.  Thinking more about this, one of the first place to apply common transformation would be filtering UI.  (We don't have filter UI yet in this new version though). 

> i was under the impression that the schema also encapsulates the data?

The schema reads the data to create a schema info (describing each field in the dataset) and calculate basic statistics, but the schema object does not store the data.  :)

PR #336, opened 2017-05-24
333 add ui tests

tafsiri:
Adds a bunch of infrastructure to run ui tests. These can be run separately from logic tests. Also adds a few initial tests.


tafsiri:
note some of the tests are WIP. i just skip them using jests api

src/components/app.test.ui.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+
+import {Provider} from 'react-redux';
+import {configureStore} from '../store';
+import {App} from './app';
+
+describe('Voyager', () => {
+
+  describe('instantiation via component', () => {
+    test('renders voyager', done => {
```
kanitw: Very minor point, but we mostly use `describe` and `it` elsewhere as it's more consistent with BDD style.  (If I understand correctly, `test` is used more with `suite` -- but yeah this is very minor as they are equivalent.)

src/components/app.test.ui.tsx
```diff
@@ -0,0 +1,85 @@
...
+                  data={data}
+                  dispatch={store.dispatch}
+              />
+            </Provider>,
+          );
+
+          const fieldList = wrapper.find('.field-list__field-list-item');
+          const fields = fieldList.children();
+
+          // expect(fieldList.get(0).text()).toBe("fieldA"); // TODO Fix;
```
kanitw: Is this the part that you mentioned as WIP?

tafsiri: oops that isn't really the part. i will update or remove this assertion

src/components/app.test.ui.tsx
tafsiri: good catch, i can update that (i tend to default to the non bdd style).

src/lib-voyager.test.ui.tsx
```diff
@@ -0,0 +1,145 @@
...
+            const fieldList = document.querySelectorAll('.field-list__field-list-item');
+            const fields = Array.prototype.map.call(fieldList, (d: Node) => d.textContent);
+
+            expect(fieldList.length).toBe(2);
+            expect(fields).toBe(['fieldA', 'fieldB']);
+            done();
+          }, 4000);
+
+          // tslint:disable-next-line:no-string-literal
+          // const _store = voyager['store'];
```
kanitw: FYI, there are some more commented lines here, not sure if they are also WIP.  (Or only the "skipped" parts are WIPs)

tafsiri: Yes these ones are part of my WIP test (currently skipped by the test runner)

PR #339, opened 2017-05-30
allow recommend to come from server url.

vlandham:
results now store plot objects as they are easily serializable.

Changes work toward accomplishing #314 .


src/actions/result.ts
```diff
@@ -18,22 +20,26 @@ export type ResultRequest = ReduxAction<typeof RESULT_REQUEST, {}>;
...
-export function resultRequest(query?: Query, schema?: Schema): AsyncResultRequest {
+export function resultRequest(query?: Query, schema?: Schema, data?: Data): AsyncResultRequest {
   return (dispatch: Dispatch<Action>, getState) => {
     if (!query) {
       query = getQuery(getState());
     }
     if (!schema) {
       schema = getSchema(getState());
     }
+    if (!data) {
```
kanitw: It seems like the input  `query`, `schema`, `data` are actually ignored and override

I wonder if we only need only one between (1) `query`, `schema`, `data` as parameters of `resultRequest`  or (2) `getState` as a parameter of the returned method.

vlandham: good point. I updated and removed them from the function signature. originally, the idea was to use this in multiple places - but based on how we are calling it now, it is not necessary. thanks!

vlandham:
hmmm. the build process seems to be having problems installing yarn

```
Updating nvm
nvm.install
2.52s$ nvm install 6
Downloading and installing node v6.10.3...
Downloading https://nodejs.org/dist/v6.10.3/node-v6.10.3-linux-x64.tar.xz...
######################################################################## 100.0%
Computing checksum with sha256sum
Checksums matched!
Now using node v6.10.3 (npm v3.10.10)
Installing yarn
30.08s$ curl -o- -L https://yarnpkg.com/install.sh | bash

bash: line 1: syntax error near unexpected token `newline'
bash: line 1: `<!DOCTYPE html>'
The command "curl -o- -L https://yarnpkg.com/install.sh | bash" failed and exited with 2 during .
```

vlandham:
getting a bad gateway error from https://yarnpkg.com/install.sh

vlandham:
Looks like yarn fixed it self. 

@kanitw  - do you know if there is a way to rerun travis - or do i need to push an update to this branch?

vlandham:
fixed @kanitw !

kanitw:
Thanks.  Last comment, I'm wondering if we should add tests for `extractPlotObjects`, `convertToPlotObjectsGroup`?

vlandham:
thanks for keeping me honest @kanitw ! Found an unnecessary input parameter to `extractPlotObjects` from the tests.

vlandham:
now includes `buildSchema` in the api - so we can run that on the server as well.

vlandham:
This tweak moves @types to dev dependencies - so they don't negatively impact `voyager-server` when it pulls in this repo.

kanitw:
LGTM.  One last comment -- I wonder if it is difficult to add tests for `fetchCompassQLRecommend` and `fetchCompassQLBuildSchema`. 

I guess it will require some mock server?  If it doesn't take too long (say ~1hr), maybe it's worth adding a test for it?  If it's gonna way longer than that, let's merge this then.

vlandham:
this latest push has a few tests for these two functions.

kanitw:
LGTM.  It looks like there is a conflict with something I have just merged.  Do you wanna rebase to the latest vy2 branch?  After that we can merge :)

vlandham:
ok ! i believe the conflicts have been resolved!

PR #340, opened 2017-05-31
325 data source config

### LINKED ISSUE #325: Update data source selector to support custom data sources
Splitting off from #313 

A few things come to mind for this:

- When an dataset is loaded via external URL, the data source selector should update to display this.
- Should the app be configurable so that the data source selector can be disabled/hidden?
- Possible support for loading multiple external data sources into the selector. (see #324)

tafsiri:
Makes showing the data source selector configurable. Closes #325


src/actions/config.ts
```diff
@@ -0,0 +1,18 @@
...
+import {ReduxAction} from './redux-action';
+
+export type ConfigAction = SetConfig;
+
+export const SET_CONFIG = 'SET_CONFIG';
+export type SetConfig = ReduxAction<typeof SET_CONFIG, {
+  config: VoyagerConfig
+}>;
+
+export function setConfig(config: VoyagerConfig): SetConfig {
```
kanitw: FYI, since there are already interface for actions like `SetConfig`, I normally don't create action creator for synchronous actions.

src/models/config.ts
```diff
@@ -0,0 +1,7 @@
+export interface VoyagerConfig {
+  showDataSourceSelector?: boolean;
```
kanitw: I think this option looks reasonable for now, but longer term we might want to think how this interacts with the `data` parameter.  (If the data is provided from external app, do we need to turn this flag on automatically, etc.)

tafsiri: I figure this is more of a libvoyager concern (the wrapper can check whether data is present and decide whether to set this to false). The default value for this if nothing is passed in is `true` and thus leaves the existing behaviour in the general case. I made it optional so that you can pass an empty object as a valid config. But we could make it non optional at some point and require/force 'voyager the app' to use at minimum the default config and  'libvoyager' to specify its desired bahaviour.

src/actions/config.ts
tafsiri: do you mean you would just dispatch the object (result of the action creator) directly from the component? I see That for undo and redo but they also didn't have a payload. 

Any examples with a payload? I guess i like the encapsulation of the action creator in case the action actually ends up being more 'steps' internally than a component is concerned with.

tafsiri:
@kanitw left some responses, let me know what you think.

src/actions/config.ts
kanitw: > do you mean you would just dispatch the object (result of the action creator) directly from the component? 

Yes

> Any examples with a payload?

Most other shelf action follow this pattern too.  

>  I guess i like the encapsulation of the action creator in case the action actually ends up being more 'steps' internally than a component is concerned with.

I think this is a valid point.  However, given the rest of the app use the object directly for action with one step, it might be better to be consistent.

src/models/config.ts
kanitw: Thanks for explanation. This sounds reasonable.  Let's keep this part as is for now.

src/actions/config.ts
tafsiri: Updated. I've removed the action creator

Issue #341, opened 2017-06-01
Have voyager css respect its container


tafsiri:
In the case where it is being embedded in another app. Voyager's css should respect the bounds of its container. It currently tends to fill the whole document body.

Related to #315.

tafsiri:
Also consider bundling the css into the js bundle so that it doesn't have to be loaded separately.

tafsiri:
closed in #342

PR #342, opened 2017-06-01
Make voyager importable

tafsiri:
This makes it so that one can import the voyager wrapper from an npm install (using the github url as a source).

Two main changes:
* The built library is now checked into the repository, and the package.json file points to the exported module as its main file. I also added some instructions to README
* The tests for the wrapper now use a different directory since /lib is checked in.


tafsiri:
Added commits to close #341

tafsiri:
ping @kanitw

.gitignore
```diff
@@ -1,5 +1,5 @@
+lib-test
 dist
-lib
```
kanitw: This means that we will always checking compiled lib file to github?  

In Vega-Lite and CompassQL, we ignore the compiled folder (`build`, which is equivalent to `lib` here), and then have a publish script that included the compiled `lib` when we publish to npm and bower (via github release).

config/webpack.lib.config.js
```diff
@@ -15,13 +15,15 @@ module.exports = {
...
   output: {
-    filename: "[name].js",
+    filename: "lib-voyager.js",
     path: path.resolve(__dirname, '../lib'),
     // Add /* filename */ comments to generated require()s in the output.
     pathinfo: true,
     // There are also additional JS chunk files if you use code splitting.
     chunkFilename: 'static/js/[name].chunk.js',
-    publicPath: '/lib/'
+    publicPath: '/lib/',
```
kanitw: Since other Vega libraries use `build/` as the compiled folder file, I wonder if we should be consistent. 

That said, other libraries are toolkit but Voyager is a UI component, so we don't really have to be the same.  
(Just asking)

tafsiri: Yeah i'm not sure what we want to do here. since there is also the built version of voyager the app which may want to go in `/build` and would drive deploys (of the application). For the moment that is why i used lib.

kanitw: I see.  Let's stick with `lib` then :)

.gitignore
tafsiri: Yes. We could do that too once we are publishing it on npm (in which case we could opt to just publish /lib). Though for now since its mainly a github based repo we would need it in the repo.

Would you want to just move to publishing to NPM and have people try it out from there? We'd need to work out a decent git tagging and npm publish workflow going for that.

kanitw: I think eventually we will have to do that as master is pretty much the development branch so better have people use from npm/bower? 

I think our deploy script in Vega-Lite and CompassQL could serve as a good start.

kanitw: So it depends if you want to work on other things first because working on the workflow script. 
(If you wanna do later, we can file an issue for now.)

tafsiri: How about we discuss this tomorrow at our meeting and figure out a plan. We have also been thinking it would be better to use master instead of vy2 at some point. But we can coordinate when we want to do this.

kanitw: Ok will merge and we can continue this conversation in https://github.com/vega/voyager/issues/350

PR #344, opened 2017-06-02
update typescript. add tsc compilation step.

vlandham:
This change bumps the typescript package version and adds a `tsc` compilation step - similar to compassql - to start preparing for the voyager project to be imported in `voyager-server`.


vlandham:
Right now, attempting to add `.d.ts` files to the compilation using `--declarations` fails in a number of places in the voyager project. Fixing this will be non-trivial. 

This latest commit provides a way to create `.d.ts` files for only a subset of the codebase. This can be rationalized as really only a portion of the codebase should ever be imported into a separate project. For now, this just builds type files for the `src/models` directory. I'm not sure if there are more directories to add to this.

Other things to consider:

* should the output of this compilation go into `dist/` (which we are using in this project already) or `build/` (which is what is used in compassql). 
* is there a cleaner way to add declarations for just a subset of files - without an additional tsconfig file?

kanitw:
> should the output of this compilation go into dist/ (which we are using in this project already) or build/ (which is what is used in compassql).

Yeah I think right now we have `dist/`, `lib/` and maybe `build/` which are all slightly different?

- `dist/` -- output folder for compiled standalone voyager
- `lib/` -- output folder for compiled voyager component lib
- `build/` -- we don't have this yet, but this could be compiled folder that other project like voyager-server can import?  

I don't really have a strong opinion, but it seems like all these different things serve different purposes so maybe we shouldn't merge them?

@tafsiri

tafsiri:
@kanitw is there a difference between `dist` and `build` as you have it? Could build be used by something like voyager-server and be the stand-alone deployable app?

cc @vlandham

vlandham:
currently there is no `build/` - but we could update the build scripts to probably accommodate this. i can investigate what effect the `outDir` has on the creation of the files currently going into `dist` and look to see if a run-time flag could be used to change the complication output dir.

vlandham:
ok @kanitw - this now outputs compiled files to the `build` dir.

tsconfig.json
```diff
@@ -1,13 +1,21 @@
 {
   "compilerOptions": {
     "outDir": "./dist/",
```
kanitw: Could you explain the reason that the main tsconfig output to `dist/` while declaration file to `build/`?

tsconfig.json
```diff
@@ -1,13 +1,21 @@
...
+    "rootDir": "./",
+    "target": "es5",
     "module": "commonjs",
+    "moduleResolution": "node",
```
kanitw: Why do we need to add so many new flags?

For example, we are not using decorators and enums? I'm not sure why we need to add flags for those. 
isolatedModules is also `false` by default.

kanitw: I would prefer to keep number of configs minimal, especially for ones that we don't really use so that it's easier to maintain.

vlandham: ok - these flags match up almost exactly with compassql - but we can remove most.

tsconfig.json
kanitw: I remember you saying there was a problem with putting them in the same folder, but I don't understand what's the problem yet.

tsconfig.json
kanitw: Oh I see.  My bad to put them in CompassQL then!

tsconfig.json
vlandham: I don't think there is a problem. this main `tsconfig` file is used for building `dist` - but let me check if the `outDir` has any bearing on that build process.

tsconfig.json
kanitw: Ok I've also removed these flags from CompassQL -- not sure why they are there at all.

PR #348, opened 2017-06-05
277 change dataset

### LINKED ISSUE #277: Changing Dataset should clean shelf and the history

tafsiri:
Clear the shelf and history when changing datasets. Closes #277

[Note this is based off of #346, so that should be merged first]

ping @kanitw 

I also would recommend that we refact datasetLoad and datasetUrlLoad into one datasetLoad function that can either take a URL or InlineData and do the right thing.

### DIFF
--- a/src/actions/dataset.ts
+++ b/src/actions/dataset.ts
@@ -2,11 +2,13 @@ import {build as buildSchema, Schema} from 'compassql/build/src/schema';
 import * as fetch from 'isomorphic-fetch';
 import {Dispatch} from 'redux';
 import {ThunkAction} from 'redux-thunk';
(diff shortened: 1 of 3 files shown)


tafsiri:
ping @kanitw

kanitw:
@tafsiri  Thanks for the ping. 

> I also would recommend that we refactor datasetLoad and datasetUrlLoad into one datasetLoad function that can either take a URL or InlineData and do the right thing.

I agree.   Do you want to do that here? (Or feel free to merge and create an issue to do it later.)

tafsiri:
@kanitw I'd prefer to do it separately, I looked at it briefly when implementing this and it may involve a change or two in some other places.

PR #355, opened 2017-06-07
Refactor datasetUrlLoad and datasetLoad into one action creator. Closes #351

### LINKED ISSUE #351: Refactor datasetLoad and datasetUrlLoad into one datasetLoad function

tafsiri:
Refactor datasetUrlLoad and datasetLoad into one action creator. Closes #351 

Ping @kanitw and @vlandham


tafsiri:
I've also added a commit that closes #277 again. The problem was the ordering of the clearing. We were clearing the history before loading the new dataset, which left some unnecessary things in the history. This now clears it after the new dataset is loaded.

PR #359, opened 2017-06-08
350 releases

vlandham:
This starts down the path of using compassql-like scripts to both provide a npm package release system as well as gh-pages deploy of `dist`. 

@kanitw i am not sure how to go about testing these scripts. do you have any suggestions?


vlandham:
it wouldnt' be required to remove the lib from the repo - when it is moved inside `build` - we _could_ check in the entire build directory. this would allow us to use the github link as a dependency and would delay our need to get voyager on npm. but it would clutter the repo a bit. 

preferences @kanitw ?

kanitw:
> @kanitw i am not sure how to go about testing these scripts. do you have any suggestions?

My terrible answer is we currently don't have test script for testing the deploy script.  We just run it and if it works, it works. :p  

> it wouldnt' be required to remove the lib from the repo - when it is moved inside build - we could check in the entire build directory. this would allow us to use the github link as a dependency and would delay our need to get voyager on npm. but it would clutter the repo a bit.

For Vega-Lite and CompassQL, we do not push `build/` to github so let's do the same. 

I think eventually we will need to rely on npm release as the main mechanism anyway, so let's just do it. (Not checking `build/` in also help us avoid inconsistencies between the actual code and `build/` folder (as we might often forget to push `build/`).

kanitw:
I'll merge this and try to publish.  I'll also add you as an owner.

Issue #354, opened 2017-06-07
Travis may not be running all tests


tafsiri:
Currently on master, the lib-voyager tests are failing however travis is not catching this. It might be because two different test suites are being executed, but probably needs some investigation.

Ping @kanitw

kanitw:
@tafsiri  Is it simply because the new ui test hasn't been added to https://github.com/vega/voyager/blob/master/.travis.yml?

tafsiri:
closed by #360

PR #360, opened 2017-06-08
add ui tests to travis config.

tafsiri:


tafsiri:
closes #354

tafsiri:
looks like the tests ran and passed.

ping @kanitw

kanitw:
The change in `.travis.yml` looks good to me, but omehow one of the test still fails.

tafsiri:
Thats interesting... one of the runs passes and the other fails but they both run the same tests. Maybe it just needs to be re-run?

kanitw:
Maybe -- I just rebased and force push -- let's see how this goes :)

tafsiri:
worked this time!

kanitw:
Yay!

PR #364, opened 2017-06-09
Pass in config from data store. tweak remote buildSchema

vlandham:
@kanitw with this and the latest voyager-server - both api calls should be functional and executable on the server. 

the only thing left is to determine how to populate `serverurl`in `config`.


kanitw:
FYI, test is still failing

Issue #262, opened 2017-02-16
Undo Button is somehow clickable from the beginning....


kanitw:
(kinda weird)

See #261 for how we implement undo

kanitw:
Done in #378

Issue #366, opened 2017-06-12
Data Selector Popup


kanitw:
Like the original Voyager 2, we should provide a popup that:

- Support loading arbitrary CSV / JSON files
- Support loading from URL

PR #378, opened 2017-06-12
Group actions so that they can be undone/redone in a single step. Rem…

### LINKED ISSUE #349: Undo/Redo leaves UI in inconsistent states from user perspective
Currently undo redo captures all actions, so if a user action has a number of intermediate (internal actions), those will be undone individually. Using [filters](https://github.com/omnidan/redux-undo#filtering-actions) or [groups](https://github.com/omnidan/redux-undo#grouping-actions) 
should help in making undo/redo apply to just user driven actions.

Any thoughts @kanitw?

tafsiri:
Group actions so that they can be undone/redone in a single step. Remove certain actions from history completely. Closes #349 

@kanitw: A few notes

- This adds a custom groupBy function to creation groups out of actions. However because various components like DatasetSelector and EncodingPane trigger RESULT_RECEIVE, and RESULT_REQUEST every time they are rendered they inject those two back into the history (and destroy `future` in the process). Thus they need to be excluded by a filter.

- I'm not sure if all the actions in https://github.com/vega/voyager/compare/349-undo-redo?expand=1#diff-0c230c2c950beb1f5987a2aed0eeeab3R61 are actually user trigger-able. Please review that list and let me know. I essentially put in all the actions I could find for now.

- Now because we have to filter out RESULT_RECEIVE, and RESULT_REQUEST, it just so happens that with this set of actions, those two were the ones causing the inconsistent state, so we could in theory remove the groupBy function. The question is whether the grouping functionality is more generally useful, or could be useful down the road. Also if we actually should remove some elements from `userAction` then this point will be moot (the grouping function will actually do something the filtering doesn't capture).

Give it a whirl and let me know what you find!

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -6,7 +6,7 @@ import HTML5Backend from 'react-dnd-html5-backend';
 import * as SplitPane from 'react-split-pane';
 
 import {Dispatch} from 'redux';
-import {StateWithHistory} from 'redux-undo';
+import {ActionCreators, StateWithHistory} from 'redux-undo';
 import {Data} from 'vega-lite/build/src/data';
 import {datasetLoad, SET_CONFIG} from '../actions';
 import {StateBase} from '../models/index';
@@ -37,6 +37,7 @@ class AppBase extends React.PureComponent<Props, {}> {
   }
 
   public componentWillMount() {
+    this.props.dispatch(ActionCreators.clearHistory());
     this.update(this.props);
   }
 
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -37,6 +37,9 @@ class AppBase extends React.PureComponent<Props, {}> {
   }
 
   public componentWillMount() {
+    // Clear history as redux-undo seems to always put the first action after
+    // an init into the history. This ensures we start with a fresh history once
+    // the app is about to start.
     this.props.dispatch(ActionCreators.clearHistory());
     this.update(this.props);
   }
--- a/src/reducers/index.ts
+++ b/src/reducers/index.ts
@@ -25,6 +25,7 @@ import {
   SHELF_SPEC_PREVIEW_DISABLE,
 } from '../actions';
 
+import {ActionType} from '../actions';
 import { State } from '../models/index';
 import {configReducer} from './config';
 import {datasetReducer} from './dataset';
@@ -43,10 +44,8 @@ function reducer(state: Readonly<StateBase>, action: Action): StateBase {
 
 /**
  * Exclude these actions from the history completely.
- *
- *
  */
-export const ACTIONS_EXCLUDED_FROM_HISTORY: string[] = [
+export const ACTIONS_EXCLUDED_FROM_HISTORY: ActionType[] = [
   // These actions are automatically re-triggered by some of the shelf components after
   // every state change. Including UNDO/REDO.
   RESULT_RECEIVE,
@@ -67,7 +66,7 @@ export const ACTIONS_EXCLUDED_FROM_HISTORY: string[] = [
  * of the preceding user action if one is available. If none is available it will be put
  * into its own group.
  */
-export const USER_ACTIONS: Object = toSet([
+export const USER_ACTIONS: ActionType[] = [
   // Dataset Actions
   DATASET_URL_REQUEST,
   // Shelf Actions,
@@ -81,7 +80,9 @@ export const USER_ACTIONS: Object = toSet([
   SHELF_SPEC_LOAD,
   SHELF_SPEC_PREVIEW,
   SHELF_SPEC_PREVIEW_DISABLE,
-]);
+];
+
+export const USER_ACTION_INDEX = toSet(USER_ACTIONS);
 
 /**
  * Actions that are to be grouped with actions that precede them.
@@ -92,12 +93,10 @@ export const USER_ACTIONS: Object = toSet([
  * DATASET_URL_RECEIVE,
  */
 
-export const GROUPED_ACTIONS: string[] = [
+export const GROUPED_ACTIONS: ActionType[] = [
   DATASET_INLINE_RECEIVE,
   DATASET_URL_RECEIVE,
 ];
-
-
 
 let _groupId = 0;
 function getNextGroupId(): number {
@@ -108,7 +107,7 @@ function getNextGroupId(): number {
 function groupAction(action: Action, currentState: State, previousHistory: StateWithHistory<State>): any {
   const currentActionType = action.type;
 
-  if (USER_ACTIONS[currentActionType]) {
+  if (USER_ACTION_INDEX[currentActionType]) {
     const nextGroupID = currentActionType + getNextGroupId();
     return nextGroupID;
   } else {
--- a/src/reducers/index.ts
+++ b/src/reducers/index.ts
@@ -7,8 +7,6 @@ import {StateBase} from '../models';
 
 import {SET_CONFIG} from '../actions/config';
 import {
-  DATASET_INLINE_RECEIVE,
-  DATASET_URL_RECEIVE,
   DATASET_URL_REQUEST,
 } from '../actions/dataset';
 import {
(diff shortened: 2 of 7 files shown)


src/reducers/index.ts
```diff
@@ -18,9 +42,64 @@ function reducer(state: Readonly<StateBase>, action: Action): StateBase {
...
+  SHELF_SPEC_PREVIEW_DISABLE,
+];
+
+let _groupId = 0;
+function getNextGroupId(): number {
+  _groupId += 1;
+  return _groupId;
+}
+
+const groupAction = (action: Action, currentState: State, previousHistory: StateWithHistory<State>): any => {
```
kanitw: (Minor) why not `function groupAction`?

src/reducers/index.ts
```diff
@@ -18,9 +42,64 @@ function reducer(state: Readonly<StateBase>, action: Action): StateBase {
   };
 }
 
+/**
+ * Exclude these actions from the history completely.
+ */
+const excludeActionsInHistory = [DATASET_INLINE_RECEIVE,
```
kanitw: A few questions

1) I understand that RESULT_RECEIVE/REQUEST is automatically triggered by spec change in the DatasetSelector and EncodingPane components, so they will get reapplied when a relevant redo is triggered.  

However, I'm not sure how `DATASET_INLINE_RECEIVE` get re-triggered.
Do we filter them because whenever new dataset load we clear all history?  
If so, we should write comments for this array describing why certain actions are excluded. (We can group actions why their reasons for exclusion.) 



2) Why is DATA_INLINE_RECEIVE here but DATASET_URL_RECEIVE not here?

3) Why should SET_CONFIG be excluded?  (Add comment)

4) (minor) line break here?  

```
const excludeActionsInHistory = [
   DATASET_INLINE_RECEIVE,
...
```

kanitw: Since this is a const array, maybe ACTIONS_EXCLUDED_FROM_HISTORY?

src/reducers/index.ts
```diff
@@ -18,9 +42,64 @@ function reducer(state: Readonly<StateBase>, action: Action): StateBase {
...
+];
+
+/**
+ * A list of actions that can be initiated by a user.
+ *
+ * Each of these will start a new 'undo group'. Non-user actions will be put into the group
+ * of the preceding user action if one is available. If none is available it will be put
+ * into its own group.
+ */
+const userActions = [
```
kanitw: Since this is a const array, maybe USER_ACTIONS?

src/reducers/index.ts
```diff
@@ -18,9 +42,64 @@ function reducer(state: Readonly<StateBase>, action: Action): StateBase {
...
+  // Dataset Actions
+  DATASET_URL_REQUEST,
+  DATASET_URL_RECEIVE,
+  DATASET_INLINE_RECEIVE,
```
kanitw: Why is DATASET_INLINE_RECEIVE both a user action and an excluded one.

src/reducers/index.ts
```diff
@@ -18,9 +42,64 @@ function reducer(state: Readonly<StateBase>, action: Action): StateBase {
...
+let _groupId = 0;
+function getNextGroupId(): number {
+  _groupId += 1;
+  return _groupId;
+}
+
+const groupAction = (action: Action, currentState: State, previousHistory: StateWithHistory<State>): any => {
+  const currentActionType = action.type;
+
+  if (userActions.indexOf(currentActionType) >= 0) {
```
kanitw: Maybe use `toSet` method in Vega-Lite's util.ts  (which in turn is imported from vega-util) to make USER_ACTION_INDEX so we can do `O(1)` check here?

tafsiri: true i was also thinking of using Array.includes or an ES6 set if typescript supports that. I will look into it! I'm probably leaning towards an ES6 set

src/reducers/index.ts
tafsiri: no reason.

src/reducers/index.ts
tafsiri: 2,3: For _DATASET_INLINE_RECEIVE_. At the moment that is not something a user action can trigger, so I don't think it should be in the history at all (Though co-incidentally  it will be wiped from the history once the action is complete anyway). Same reason for SET_CONFIG (minus the auto clearing of history). I can add a comment to this effect. However if you can think of reasons why they should be in the history, let me know.

4. +1 on line break

src/reducers/index.ts
tafsiri: That is an oversight, it probably shouldn't be here. Again unless we can think of a reason why it should be here. At the moment i just can't think of one.

src/reducers/index.ts
tafsiri: Sure, I can go with that convention.

tafsiri:
@kanitw i'm making the requested changes, but let me know if you saw any actions in USER_ACTIONS that should be grouped with an existing action (but still remain in the history). As mentioned before currently the exclude actions takes care of it all at the moment, but maybe the grouping mechanism is still useful down the road?

tafsiri:
FYI when this lands. you will need to update your dependencies. cc @vlandham

kanitw:
LGTM except one last question, I still do not understand why is DATASET_INLINE_RECEIVE treated differently from DATASET_URL_RECEIVE.

tafsiri:
Ah, because the UI never triggers DATASET_INLINE_RECEIVE. This is the method triggered by update data if inline data is passed in, but from what i can tell, that can't happen from the UI (ui only uses urlData). 

So when embedding voyager one can trigger this action, however i don't know if you would then want the user to be able to undo that? What do you think?

kanitw:
> Ah, because the UI never triggers DATASET_INLINE_RECEIVE. This is the method triggered by update data if inline data is passed in, but from what i can tell, that can't happen from the UI (ui only uses urlData).

This might not be true in the near future.  Once we re-add an ability to load inline data in the data selector popup (#366), inline data may be triggered from the UI.

tafsiri:
@kanitw then i would say we can remove it from the list at that time :)

kanitw:
Actually now I wonder why are DATASET_URL_REQUEST/RECEIVE in the grouped list at all -- since when we change dataset, we clear the history anyway, so users can't really undo dataset url requests/ receive.

tafsiri:
@kanitw, its in the list because it shouldn't be grouped with some other action that is in the list. Given the effect of the action creator we could probably remove it, but its kind of a side effect. I don't think it matters too much either way, but i like the clarity of it being here. 

I'm not sure if it would make it harder to test in isolation. I could go either way on this one (though DATASET_URL_REQUEST does not clear the history so we should leave that in).

tafsiri:
@kanitw updated. You convinced me on some of the action placement, so i moved some around. I put a comment block for some of the grouped actions.

tafsiri:
Also closes #262

src/components/app.tsx
```diff
@@ -37,6 +37,7 @@ class AppBase extends React.PureComponent<Props, {}> {
   }
 
   public componentWillMount() {
+    this.props.dispatch(ActionCreators.clearHistory());
```
kanitw: Maybe add comment that this is a workaround?

tafsiri:
added test to check actions are assigned to groups specifying how they interact with the history

src/reducers/index.ts
```diff
@@ -18,9 +42,84 @@ function reducer(state: Readonly<StateBase>, action: Action): StateBase {
   };
 }
 
+/**
+ * Exclude these actions from the history completely.
+ */
+export const ACTIONS_EXCLUDED_FROM_HISTORY: ActionType[] = [
```
kanitw: Actually, since `ACTIONS_EXCLUDED_FROM_HISTORY`, `USER_ACTIONS`, and `GROUPED_ACTIONS` are declarations about actions, maybe they should be in `src/actions/index.ts` more?

kanitw: ps.  If you agree, we should move the tests too.

tafsiri: No... i think these are declarations about history rather than actions.

kanitw: Hmm. I think they are declarations of different categories of actions.   The purpose is for history, but whenever you work on new action, you have to add each action to one of these arrays.

tafsiri: Exactly, the purpose is for history; absent of history functionality that information isn't used, so my recommendation is to keep it close to how/where history is configured. Your call, but thats how it seems to me at the moment.

kanitw: Let's keep it as is for now then.  We can move later if needed.

PR #383, opened 2017-06-15
338 table schema

vlandham:
For #338 - update Voyager to latest version of compassql so as to support table schema.


kanitw:
Thanks!

PR #386, opened 2017-06-16
Add SERVER env variable for setting serverUrl.

vlandham:
closes #376

PR #387, opened 2017-06-16
Update README.md

kanitw:


kanitw:
@vlandham @tafsiri  -- Please help add missing information.  

A few ones I can think of:

- The new start via server in https://github.com/vega/voyager/pull/386 (Do we have to run voyager-server separately?) 
- Describe the different output folders we produce

tafsiri:
also @vlandham could you update the package.json main entry to point to the new build dir. `"main": "build/lib-voyager.js",`

PR #393, opened 2017-06-19
Cyn/header ui

starry97:
fixed header css


src/components/header/controls.tsx
```diff
@@ -0,0 +1,18 @@
...
+import * as CSSModules from 'react-css-modules';
+import { Bookmark } from './bookmark';
+import * as styles from './controls.scss';
+import { UndoRedo } from './undo-redo';
+
+class ControlsBase extends React.Component<any, any> {
+  public render() {
+    return (
+      <div styleName='controls'>
+        <Bookmark/>
```
kanitw: Let's comment out this line for now so we don't include a button that does not work :)

src/components/header/bookmark.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+
+export class Bookmark extends React.PureComponent<any, any> {
```
kanitw: wonder if replacing any with {} works here.  If it does, that's better

src/components/header/controls.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import * as CSSModules from 'react-css-modules';
+// import { Bookmark } from './bookmark';
+import * as styles from './controls.scss';
+import { UndoRedo } from './undo-redo';
+
+class ControlsBase extends React.Component<any, any> {
```
kanitw: wonder if replacing any with {} works here.  If it does, that's better

kanitw:
@starry97 -- FYI, I further converted any => {} for you.  I'll merge since this looks good otherwise. :)

PR #402, opened 2017-06-22
Data Modal

### LINKED ISSUE #366: Data Selector Popup

vlandham:
This works to close #366 

it is not very pretty yet, but it works!

data modal

One problem: the `@types` for `react-modal` is out of date, so I was trying to fake the declarations to be able to use the component. I've tried half a dozen variants, but the one that is checked in has:

```
declare module 'react-modal' {
  const Modal: any;
  export default Modal;
}
import Modal from 'react-modal';
```

in the data-selector

This was derived from instructions on http://www.typescriptlang.org/docs/handbook/modules.html

However, this doesn't work. I get the errors:

screen shot 2017-06-22 at 3 25 27 pm

I'm doing something very foolish with Typescript - but I don't know what it is. Help!


src/components/data-pane/data-selector.scss
```diff
@@ -0,0 +1,179 @@
...
+.modal {
+  display: flex;
+  margin: 50px auto;
+  width: calc( 100% - 100px);
+  padding: 6px;
+  background-color: #E2E9F3;
+  border: 1px solid #aaa;
+  border-radius: 3px;
+  box-sizing: border-box;
+  display: -webkit-box;
```
kanitw: I think we have `autoprefixer` so do not have to do this manually.

PR #405, opened 2017-06-23
Rename methods for mutating immutable arrays

kanitw:
@starry97 
These are straightforward renaming, but I'll assign it for you to review since I think it is worth for you look at the pattern of how we use these methods.


kanitw:
(You can merge if you're a reviewer - but I just merged it anyway :)

PR #406, opened 2017-06-23
Make app state serializable 

### LINKED ISSUE #398: App state not serializable
In the course of working on https://github.com/vega/voyager-electron/pull/9, it appears that the voyager app state (redux store state) is not serializable. Making an issue here to discuss what would be involved in making it serializable. 

Here are some of the errors i get when trying to load state that has been run through JSON.stringify.  I'll also attach an example state file.

They are unfortunately difficult to trace from the built code that voyager-electron uses.

screen shot 2017-06-20 at 3 18 02 pm

screen shot 2017-06-20 at 3 18 18 pm

Link to gist with sample serialized state: https://gist.github.com/tafsiri/c6a6ef84379ed39c7777df8a06a68fb1

So what do folks already know about schema.cardinality and other parts of the state tree with respect to serialization?

Ping @vlandham @kanitw

tafsiri:
Make app state serializable. Add helper methods to convert from internal representation to serializable form. Closes #398

pint @kanitw @vlandham

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -31,16 +31,6 @@ export function fetchCompassQLRecommend(query: Query, schema: Schema, data: Data
       response => {
         return response.json();
       }
-    ).then(
-      fields => {
-        return {
-          name: fields.name,
-          path: fields.path,
-          items: fields.items,
-          groupBy: fields.groupBy,
-          orderGroupBy: fields.orderGroupBy,
-        };
-      }
     );
   } else {
     return new Promise(resolve => {
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -33,8 +33,13 @@ export function fetchCompassQLRecommend(query: Query, schema: Schema, data: Data
       }
     ).then(
       fields => {
-        return new SpecQueryGroup<PlotObject>(fields.name, fields.path,
-          fields.items, fields.groupBy, fields.orderGroupBy);
+        return {
+          name: fields.name,
+          path: fields.path,
+          items: fields.items,
+          groupBy: fields.groupBy,
+          orderGroupBy: fields.orderGroupBy,
+        };
       }
     );
   } else {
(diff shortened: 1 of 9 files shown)


src/api/api.ts
```diff
@@ -33,8 +33,13 @@ export function fetchCompassQLRecommend(query: Query, schema: Schema, data: Data
       }
     ).then(
       fields => {
-        return new SpecQueryGroup<PlotObject>(fields.name, fields.path,
-          fields.items, fields.groupBy, fields.orderGroupBy);
+        return {
```
kanitw: This seems to already be a object of the right type.  Maybe cut the last chain?

src/models/index.ts
```diff
@@ -30,3 +33,42 @@ export const DEFAULT_STATE: StateBase = {
...
+    },
+    tableschema: state.dataset.schema.tableSchema(),
+  };
+
+  return asSerializable;
+}
+
+export function fromSerializable(serializable: SerializableState): Readonly<StateBase> {
+  // We make a clone of this object to not modify the input param.
+  const deserialized = JSON.parse(JSON.stringify(serializable));
```
kanitw: We have `duplicate` in vega-lite's `util` which call parse & stringify, but better use consistent calls?

src/models/index.ts
```diff
@@ -30,3 +33,42 @@ export const DEFAULT_STATE: StateBase = {
   shelf: DEFAULT_SHELF_SPEC,
   result: DEFAULT_RESULT,
 };
+
+export interface SerializableState {
```
kanitw:
```
export type SerializableState = Pick<StateBase, 'config' | 'shelf' | 'result' | 'dataset'> & {
  tableschema: TableSchema<FieldSchema>;
}
```
would be less typing when we modify `StateBase` in the future

src/api/api.ts
tafsiri: @vlandham is this always going to return an object of this shape.

vlandham: i believe so yes.

Issue #353, opened 2017-06-07
Redesign copy button / export mechanism


kanitw:
## Basic

- [x] Basic Copy Vega-Lite Spec
- [x] Replace the copy button with a nicer icon

kanitw:
@starry97  I'm adding 

- [ ] Replace the copy button with a nicer icon 

above :)

kanitw:
Re-design copy button so that it works with data binding

kanitw:
Inline Data
- Include inline data
- Named data source

URL Data
- URL Data
- Named data source
- Include inline data -- least important ...

kanitw:
For wrapper projects (e.g., Jupyter), we may want other export mechanism instead too.

PR #410, opened 2017-06-27
click a button to copy spec

starry97:
part of #353


kanitw:
Thanks!  This works for now. In the near future, we should replace this with some nicer icons.

PR #411, opened 2017-06-27
added type changer ui

### LINKED ISSUE #368: Field Type Changer
voyager_2

starry97:
fix #368

### DIFF
--- a/src/components/data-pane/field-list.tsx
+++ b/src/components/data-pane/field-list.tsx
@@ -40,17 +40,14 @@ class FieldListBase extends React.PureComponent<FieldListProps, FieldListState> 
 
   public render() {
     const {fieldDefs, schema} = this.props;
-    const fieldItems = [];
-    for (let i = 0; i < fieldDefs.length; i++) {
-      const fieldDef = fieldDefs[i];
-      const fieldSchemas = schema.fieldSchemas;
-      let type: PrimitiveType;
-      if (fieldSchemas && fieldSchemas[i]) {
-        type = fieldSchemas[i].type;
+    const fieldItems = fieldDefs.map(fieldDef => {
+      let primitiveType;
+      if (typeof fieldDef.field === 'string') {
+        primitiveType = schema.primitiveType(fieldDef.field);
       }
-      const hideTypeChanger = (type !== PrimitiveType.NUMBER && type !== PrimitiveType.INTEGER)
-        || fieldDef.field === '?';
-      fieldItems.push (
+      const hideTypeChanger = (primitiveType !== PrimitiveType.NUMBER && primitiveType !== PrimitiveType.INTEGER)
+        || typeof fieldDef.field !== 'string';
+      return (
         <div key={JSON.stringify(fieldDef)} styleName="field-list-item">
           <TetherComponent
              attachment="top left"
@@ -65,11 +62,10 @@ class FieldListBase extends React.PureComponent<FieldListProps, FieldListState> 
               onDoubleClick={this.onAdd}
               onAdd={this.onAdd}
             />
-            {!hideTypeChanger && this.renderTypeChanger(fieldDef, type)}
+            {!hideTypeChanger && this.renderTypeChanger(fieldDef, primitiveType)}
           </TetherComponent>
       </div>);
-    }
-
+    });
     return (
       <div className="FieldList">
         {fieldItems}
--- a/src/components/data-pane/field-list.tsx
+++ b/src/components/data-pane/field-list.tsx
@@ -5,7 +5,7 @@ import * as TetherComponent from 'react-tether';
 import * as styles from './field-list.scss';
 
 import {ExpandedType} from 'compassql/build/src/query/expandedtype';
-import {PrimitiveType, Schema, FieldSchema} from 'compassql/build/src/schema';
+import {PrimitiveType, Schema} from 'compassql/build/src/schema';
 import {DatasetSchemaChangeFieldType} from '../../actions/dataset';
 import {ActionHandler, createDispatchHandler} from '../../actions/redux-action';
 import {SHELF_FIELD_AUTO_ADD, ShelfFieldAutoAdd} from '../../actions/shelf';
--- a/src/components/data-pane/field-list.tsx
+++ b/src/components/data-pane/field-list.tsx
@@ -1,8 +1,12 @@
 import * as React from 'react';
 import * as CSSModules from 'react-css-modules';
 import {connect} from 'react-redux';
+import * as TetherComponent from 'react-tether';
 import * as styles from './field-list.scss';
 
+import { ExpandedType } from 'compassql/build/src/query/expandedtype';
+import { PrimitiveType, Schema } from 'compassql/build/src/schema';
+import { DatasetSchemaChangeFieldType } from '../../actions/dataset';
 import {ActionHandler, createDispatchHandler} from '../../actions/redux-action';
 import {SHELF_FIELD_AUTO_ADD, ShelfFieldAutoAdd} from '../../actions/shelf';
 import {FieldParentType} from '../../constants';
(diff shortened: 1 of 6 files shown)


src/components/data-pane/type-changer.tsx
```diff
@@ -0,0 +1,62 @@
...
+    this.state = {
+      selectedType: this.props.fieldDef.type
+    };
+  }
+  public render() {
+    const {types} = this.props;
+    return (
+      <div styleName='type-changer'>
+        <h4>Type</h4>
+        {types.map(ele => {
```
kanitw: Rename `ele` => `type` ?

src/components/data-pane/field-list.tsx
```diff
@@ -1,48 +1,78 @@
...
   }
 
   public render() {
-    const {fieldDefs} = this.props;
-
-    const fieldItems = fieldDefs.map(fieldDef => {
-      return (
+    const {fieldDefs, handleAction, schema} = this.props;
+    const fieldItems = [];
+    for (let i = 0; i < fieldDefs.length; i++) {
```
kanitw: I'm curious why you replace map with for loop here.  
(`map` seems cleaner?)

src/components/data-pane/type-changer.tsx
```diff
@@ -0,0 +1,62 @@
...
+import * as CSSModules from 'react-css-modules';
+
+import { ExpandedType } from 'compassql/build/src/query/expandedtype';
+import {DATASET_SCHEMA_CHANGE_FIELD_TYPE, DatasetSchemaChangeFieldType} from '../../actions/dataset';
+import {ActionHandler} from '../../actions/redux-action';
+import {ShelfFieldDef} from '../../models/shelf/encoding';
+import * as styles from './type-changer.scss';
+
+export interface TypeChangerProps extends ActionHandler<DatasetSchemaChangeFieldType> {
+  fieldDef: ShelfFieldDef;
```
kanitw: Instead of require `ShelfFieldDef` in which `field` can be either string or wildcard.  Let's not make `fieldDef` a prop and instead have `field: string` and `type` directly (since we're not using other parameters anyway.  :)

src/components/data-pane/type-changer.tsx
```diff
@@ -0,0 +1,62 @@
...
+  types: ExpandedType[];
+}
+
+export interface TypeChangerState {
```
kanitw: I wonder if we really need internal state for selectedType here.  
If I understand correctly, I think we should not have it as we can directly read the input `type` props.

src/components/data-pane/type-changer.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <div styleName='type-changer'>
+        <h4>Type</h4>
+        {types.map(ele => {
+          return (
+            <label key={ele}>
+              <input type='radio' value={ele} name='type'
+                onChange={this.onTypeChange.bind(this)}
+                checked={this.state.selectedType === ele}
+              />
+              <span styleName='type'> {ele} </span>
```
kanitw: Is the space between span tag and `{ele}` intended?

src/components/data-pane/field-list.tsx
```diff
@@ -1,48 +1,78 @@
...
-            draggable={true}
-            parentId={{type: FieldParentType.FIELD_LIST}}
-
-            onDoubleClick={this.onAdd}
-            onAdd={this.onAdd}
-          />
-        </div>
-      );
-    });
+          <TetherComponent
```
kanitw: Let's only initiate Tether with TypeChanger only if `fieldDef.type` is a `string`.  (You can extract this to a method so `render()` is not too complicated.

src/components/data-pane/field-list.tsx
```diff
@@ -1,48 +1,78 @@
 import * as React from 'react';
 import * as CSSModules from 'react-css-modules';
 import {connect} from 'react-redux';
+import * as TetherComponent from 'react-tether';
 import * as styles from './field-list.scss';
 
+import { ExpandedType } from 'compassql/build/src/query/expandedtype';
```
kanitw: Since it would take 10 sec to fix this let's make it consistent :p

src/components/data-pane/type-changer.tsx
starry97: Yes, just to add some margin between the options.

src/components/data-pane/field-list.tsx
starry97: Because I want to use indexes to get the schema

src/components/data-pane/field-list.tsx
```diff
@@ -1,48 +1,74 @@
...
-    const {fieldDefs} = this.props;
-
-    const fieldItems = fieldDefs.map(fieldDef => {
-      return (
+    const {fieldDefs, schema} = this.props;
+    const fieldItems = [];
+    for (let i = 0; i < fieldDefs.length; i++) {
+      const fieldDef = fieldDefs[i];
+      const fieldSchemas = schema.fieldSchemas;
+      let type: PrimitiveType;
```
kanitw: Let's name this variable `primitiveType` so that `type` in our codebase always mean `ExpandedType`.  

Also, I think we have `schema.primitiveType(field)` so you don't really need to get the fieldSchema directly.

src/components/data-pane/field-list.tsx
```diff
@@ -1,49 +1,71 @@
...
-            isPill={true}
-            draggable={true}
-            parentId={{type: FieldParentType.FIELD_LIST}}
-
-            onDoubleClick={this.onAdd}
-            onAdd={this.onAdd}
-          />
-        </div>
-      );
+          <TetherComponent
```
kanitw: if `hideTypeChanger` is true, why won't we just render the `Field`  -- no need for TetherComponent is such case?

src/components/data-pane/type-changer.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import {ExpandedType} from 'compassql/build/src/query/expandedtype';
+import {DATASET_SCHEMA_CHANGE_FIELD_TYPE, DatasetSchemaChangeFieldType} from '../../actions/dataset';
+import {ActionHandler} from '../../actions/redux-action';
+import * as styles from './type-changer.scss';
+
+export interface TypeChangerProps extends ActionHandler<DatasetSchemaChangeFieldType> {
+  field: string;
+  type: ExpandedType;
+  types: ExpandedType[];
```
kanitw: maybe `types` => `validTypes` ?

src/components/data-pane/index.tsx
```diff
@@ -5,12 +5,14 @@ import {connect} from 'react-redux';
 import * as styles from './data-pane.scss';
 
 import {ActionHandler, createDispatchHandler, DatasetAsyncAction, ShelfFieldAutoAdd} from '../../actions';
+import { DatasetSchemaChangeFieldType } from '../../actions/dataset';
```
kanitw: remove space?

src/components/data-pane/field-list.tsx
```diff
@@ -76,7 +133,8 @@ export const FieldList = connect(
 export const PresetWildcardFieldList = connect(
   (state: State) => {
     return {
-      fieldDefs: getPresetWildcardFields(state)
+      fieldDefs: getPresetWildcardFields(state),
+      schema: state.present.dataset.schema
```
kanitw: Should we use the selector for `getSchema` here?  I guess we should for consistency

src/components/data-pane/field-list.tsx
```diff
@@ -58,6 +80,40 @@ class FieldListBase extends React.PureComponent<FieldListProps, {}> {
...
+      });
+    } else {
+      this.setState({
+        selectedField: field
+      });
+    }
+  }
+
+  private getTypes(primitiveType: PrimitiveType): ExpandedType[] {
+    if (primitiveType === PrimitiveType.NUMBER) {
```
kanitw: Let's do a `switch` and correctly determine types for all primitive types?

(string = NOMINAL only, datetime = TEMPORAL only, etc.)

src/components/data-pane/field-list.tsx
```diff
@@ -1,49 +1,71 @@
...
   public render() {
-    const {fieldDefs} = this.props;
-
+    const {fieldDefs, schema} = this.props;
     const fieldItems = fieldDefs.map(fieldDef => {
+      let primitiveType;
+      if (typeof fieldDef.field === 'string') {
+        primitiveType = schema.primitiveType(fieldDef.field);
+      }
+      const hideTypeChanger = (primitiveType !== PrimitiveType.NUMBER && primitiveType !== PrimitiveType.INTEGER)
```
kanitw: Instead of hard coding primitiveType check here, maybe call `getTypes` and check if there are more than one valid types (which mean we can change?).

src/components/data-pane/field-list.tsx
```diff
@@ -58,6 +80,40 @@ class FieldListBase extends React.PureComponent<FieldListProps, {}> {
...
+        selectedField: null
+      });
+    } else {
+      this.setState({
+        selectedField: field
+      });
+    }
+  }
+
+  private getTypes(primitiveType: PrimitiveType): ExpandedType[] {
```
kanitw: `getValidTypes`?

src/components/data-pane/field-list.tsx
```diff
@@ -1,49 +1,57 @@
 import * as React from 'react';
 import * as CSSModules from 'react-css-modules';
 import {connect} from 'react-redux';
+import * as TetherComponent from 'react-tether';
 import * as styles from './field-list.scss';
 
+import {ExpandedType} from 'compassql/build/src/query/expandedtype';
+import {PrimitiveType, Schema} from 'compassql/build/src/schema';
+import {ShelfFieldDef} from '../../../build/src/models/shelf/encoding';
```
kanitw: This should be `../../models/shelf/encoding` (we should import from original source, not the compiled ones.)  

I'll add 
```
"tsimporter.filesToExclude": ["build/**"],
```

to `master`

src/components/data-pane/field-list.tsx
starry97: In `src/components/data-pane/index.tsx`, we did not use the selector.

PR #414, opened 2017-06-29
Add DatasetSchemaChangeOrdinalDomain action and reducer

felixcodes:
Final part of https://github.com/vega/compassql/issues/87

Waiting for https://github.com/vega/compassql/pull/394 PR to be reviewed and merged in. Otherwise this test will fail.


src/reducers/dataset.test.ts
```diff
@@ -134,4 +134,57 @@ describe('reducers/dataset', () => {
...
+      };
+
+      const simpleDataset: Dataset = {
+        isLoading: false,
+        name: 'Test',
+        schema: new Schema({fields:
+        [{
+          name: 'o',
+          vlType: 'ordinal',
+          type: 'string' as any,
```
kanitw: not sure why we need `as any` here

src/reducers/dataset.test.ts
```diff
@@ -134,4 +134,57 @@ describe('reducers/dataset', () => {
...
+
+        data: data
+      };
+
+      const changedSchema = new Schema({fields:
+        [{
+          name: 'o',
+          ordinalDomain: ['B', 'A'],
+          vlType: 'ordinal',
+          type: 'string' as any,
```
kanitw: not sure why we need `as any` here

src/reducers/dataset.ts
```diff
@@ -12,57 +13,62 @@ import {FieldSchema, Schema} from 'compassql/build/src/schema';
...
   return dataset;
 }
 
-export function changeFieldType(schema: Schema, field: string, type: ExpandedType) {
-  const changedFieldSchema: FieldSchema = {
-    ...schema.fieldSchema(field),
-    vlType: type
-  };
-
+function updateSchema(schema: Schema, changedFieldSchema: FieldSchema, field: string) {
```
kanitw: stylistically, I would make changedFieldSchema the last parameter so that you don't need to declare `const changedFieldSchema` in dependent methods.

src/reducers/dataset.test.ts
felixcodes: note: We agreed that this is due to Typescript and is OK.

Issue #373, opened 2017-06-12
Filter UI 


kanitw:
- [x] Read about filter in Vega Lite

  - [x] We won't use Vega expression directly

- [x] Update CompassQL -- to support it  (https://github.com/vega/compassql/issues/366)

- [x] Design State and reducer
  - Although Filter is a part of `transform` *array*  in Vega-Lite (and will be in CompassQL), I don't think we need it to be an array in Voyager (for now). 

-------
- [x] Look at both Tableau and the old UI

- [x] Support Drag and Drop
- [ ] Filter for Quantitative
  - [x] Support entering specific value (Rather than just adjusting range)
  - [ ] Determine the right filter step
- [x] Filter for Nominal + Search Box
  - [x] Clear all
- [x] Filter for Temporal and Time Unit

starry97:
Open #534 for filter step

PR #415, opened 2017-06-29
filter action and reducer

starry97:
part of #373


src/components/view-pane/index.tsx
```diff
@@ -4,26 +4,27 @@ import * as React from 'react';
...
-import {getData, getMainResult, getQuery} from '../../selectors';
+import {getData, getFilters, getMainResult, getQuery} from '../../selectors';
 import {Plot} from '../plot';
 import {PlotList} from '../plot-list';
 import * as styles from './view-pane.scss';
 
 export interface ViewPaneProps extends ActionHandler<ShelfAction> {
   data: Data;
   query: Query;
+  filters: Transform[];
```
kanitw: Instead of all possible transform, this could be just a subset of transform that we support.

src/models/plot.ts
```diff
@@ -33,7 +33,8 @@ export function extractPlotObjects(modelGroup: SpecQueryGroup<PlotObject>) {
   });
 }
 
-export function convertToPlotObjectsGroup(modelGroup: SpecQueryModelGroup, data: Data): SpecQueryGroup<PlotObject> {
+export function convertToPlotObjectsGroup(modelGroup: SpecQueryModelGroup,
```
kanitw: I like this tabbing style better

```
export function convertToPlotObjectsGroup(
  modelGroup: SpecQueryModelGroup,
  data: Data
): SpecQueryGroup<PlotObject> {
```

src/models/shelf/spec.ts
```diff
@@ -27,6 +25,8 @@ export interface ShelfUnitSpec {
   anyEncodings: ShelfAnyEncodingDef[];
 
   config: Config;
+
+  filters?: Transform[];
```
kanitw: Correct type of filters to be more specific

src/actions/filter.ts
```diff
@@ -0,0 +1,26 @@
+import {FilterTransform} from 'vega-lite/build/src/transform';
+import {ReduxAction} from './redux-action';
+
+export type FilterAction = AddRangeFilter | AddOneOfFilter | RemoveRangeFilter | RemoveOneOfFilter ;
+
+export const ADD_RANGE_FILTER = 'ADD_RANGE_FILTER';
```
kanitw: For consistency, let's name this `FILTER_ADD_RANGE`.    (Use the part of state modified as prefix)

src/actions/filter.ts
```diff
@@ -0,0 +1,26 @@
+import {FilterTransform} from 'vega-lite/build/src/transform';
+import {ReduxAction} from './redux-action';
+
+export type FilterAction = AddRangeFilter | AddOneOfFilter | RemoveRangeFilter | RemoveOneOfFilter ;
```
kanitw: Besides Adding / Removing, you will have to support modifying range / oneof filter as well.   
(But I'm fine if the modifying parts come in another PR, which will make it easier to review.)

kanitw: Tip: You can reuse the array utility I have

src/reducers/shelf/filter.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    case ADD_ONE_OF_FILTER: {
+      const {filter} = action.payload;
+      const filters = insertItemToArray(shelfSpec.filters, 0, filter);
+      return {
+        ...shelfSpec,
+        filters: filters
+      };
+    }
+    case REMOVE_RANGE_FILTER: {
```
kanitw: I think we can have just one `FILTER_REMOVE`  that takes `index` as parameter and remove the item from the array.  (And no need to do O(N) search.)

src/reducers/shelf/filter.ts
```diff
@@ -0,0 +1,47 @@
+
+import {ADD_ONE_OF_FILTER, ADD_RANGE_FILTER, REMOVE_ONE_OF_FILTER, REMOVE_RANGE_FILTER} from '../../actions/filter';
+import {Action} from '../../actions/index';
+import {DEFAULT_SHELF_UNIT_SPEC, ShelfUnitSpec} from '../../models/shelf/spec';
+import {insertItemToArray, removeItemFromArray} from './spec';
+
+export function filterReducer(shelfSpec: Readonly<ShelfUnitSpec> = DEFAULT_SHELF_UNIT_SPEC,
+                              action: Action): ShelfUnitSpec {
+  switch (action.type) {
+    case ADD_RANGE_FILTER: {
```
kanitw: These two ADD do exactly the same thing, it seems like we only need one `FILTER_ADD`?

src/reducers/shelf/spec.ts
```diff
@@ -215,7 +215,7 @@ function removeEncoding(shelf: Readonly<ShelfUnitSpec>, shelfId: ShelfId):
 /**
  * Immutable array splice
  */
-function removeItemFromArray(array: ReadonlyArray<any>, index: number) {
+export function removeItemFromArray(array: ReadonlyArray<any>, index: number) {
```
kanitw: We should move these utilities to `src/reducers/util.ts`.  :)

src/selectors/index.ts
```diff
@@ -9,6 +9,7 @@ export const getShelf = (state: State) => state.present.shelf;
 export const getSchema = (state: State) => state.present.dataset.schema;
 export const getMainResult = (state: State) => state.present.result.main.modelGroup;
 export const getConfig = (state: State) => state.present.config;
+export const getFilters = (state: State) => state.present.shelf.spec.filters;
```
kanitw: Let's sort these simple reducers by alphabetical order.

src/models/shelf/spec.ts
starry97: Should filters be an optional parameter? or should it be an empty array by default?

kanitw: > Should filters be an optional parameter? or should it be an empty array by default?

I don't have strong opinion about this, but I guess empty array would lead to less code (no need to check if something is defined.)

src/models/shelf/index.test.ts
```diff
@@ -10,7 +10,8 @@ describe('models/shelf', () => {
             x: {field: 'a', type: 'quantitative'}
           },
           anyEncodings: [],
-          config: {numberFormat: 'd'}
+          config: {numberFormat: 'd'},
+          filters: []
```
kanitw: Can you move `filters` and `transform` to be above `mark` and `encoding` for all of these?  

(Since `filters` and `transform` are applied before we encode data.)

src/models/shelf/spec.ts
```diff
@@ -27,28 +26,31 @@ export interface ShelfUnitSpec {
   anyEncodings: ShelfAnyEncodingDef[];
 
   config: Config;
+
+  filters: Array<RangeFilter | OneOfFilter>;
```
kanitw: Can you move `filters` and `transform` to be above `mark` and `encoding` for all of these?

src/reducers/index.ts
```diff
@@ -89,6 +91,9 @@ export const USER_ACTIONS: ActionType[] = [
   SHELF_SPEC_LOAD,
   SHELF_SPEC_PREVIEW,
   SHELF_SPEC_PREVIEW_DISABLE,
+
+  FILTER_ADD,
```
kanitw: Add `// Filter Actions` and move to above shelf (alphabetical order).

src/reducers/shelf/filter.ts
```diff
@@ -0,0 +1,29 @@
...
+import {Action} from '../../actions/index';
+import {DEFAULT_SHELF_UNIT_SPEC, ShelfUnitSpec} from '../../models/shelf/spec';
+import {insertItemToArray, removeItemFromArray} from '../util';
+
+export function filterReducer(shelfSpec: Readonly<ShelfUnitSpec> = DEFAULT_SHELF_UNIT_SPEC,
+                              action: Action): ShelfUnitSpec {
+  switch (action.type) {
+    case FILTER_ADD: {
+      const {filter} = action.payload;
+      const filters = insertItemToArray(shelfSpec.filters, 0, filter);
```
kanitw: `FILTER_ADD` should also take index as input (so that user can drag and drop field at a specific location). 

Or at least, we should make index = length of array (so it is appended at the end, not the beginning.)

src/selectors/index.ts
```diff
@@ -4,11 +4,12 @@ import {createSelector} from 'reselect';
 
 import {Shelf, ShelfFieldDef, State, toQuery} from '../models';
 
+export const getConfig = (state: State) => state.present.config;
 export const getData = (state: State) => state.present.dataset.data;
 export const getShelf = (state: State) => state.present.shelf;
 export const getSchema = (state: State) => state.present.dataset.schema;
+export const getFilters = (state: State) => state.present.shelf.spec.filters;
```
kanitw: Alphabetical order?

kanitw:
test failing

PR #420, opened 2017-07-07
show warnings from vega lite

starry97:
screen shot 2017-07-07 at 2 17 00 pm

### DIFF
--- a/src/components/vega-lite/index.tsx
+++ b/src/components/vega-lite/index.tsx
@@ -3,11 +3,14 @@ import * as vega from 'vega';
 import * as vl from 'vega-lite';
 import {TopLevelExtendedSpec} from 'vega-lite/build/src/spec';
 import * as vegaTooltip from 'vega-tooltip';
+import {Logger} from '../../models/shelf/logger';
 
 export interface VegaLiteProps {
   spec: TopLevelExtendedSpec;
 
   renderer?: 'svg' | 'canvas';
+
+  logger: Logger;
 }
 
 const CHART_REF = 'chart';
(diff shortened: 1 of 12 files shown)


src/reducers/shelf/index.ts
```diff
@@ -2,14 +2,15 @@ import {Schema} from 'compassql/build/src/schema';
...
+import {Logger} from '../../models/shelf/logger';
 import {shelfSpecReducer} from './spec';
 import {shelfSpecPreviewReducer} from './spec-preview';
 
 export function shelfReducer(shelf: Readonly<Shelf> = DEFAULT_SHELF_SPEC, action: Action, schema: Schema): Shelf {
   return {
     spec: shelfSpecReducer(shelf.spec, action, schema),
-    specPreview: shelfSpecPreviewReducer(shelf.specPreview, action)
+    specPreview: shelfSpecPreviewReducer(shelf.specPreview, action),
+    logger: new Logger()
```
kanitw: Why do we need to create a new logger every time?

src/components/vega-lite/index.tsx
```diff
@@ -3,27 +3,44 @@ import * as vega from 'vega';
...
+  warnings: any[];
 }
 
 const CHART_REF = 'chart';
 
-export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
+export class VegaLite extends React.PureComponent<VegaLiteProps, VegaLiteState> {
+  constructor(props: VegaLiteProps) {
+    super(props);
+    this.state = {
```
kanitw: Why do we need to use local state?

src/components/vega-lite/index.tsx
```diff
@@ -3,27 +3,44 @@ import * as vega from 'vega';
...
 import * as vegaTooltip from 'vega-tooltip';
+import {Logger} from '../../models/shelf/logger';
+import {Warnings} from './warnings';
 
 export interface VegaLiteProps {
   spec: TopLevelExtendedSpec;
 
   renderer?: 'svg' | 'canvas';
+
+  logger: Logger;
```
kanitw: Currently logger is a singleton logger stored in the global redux store, but it is render in all VegaLite plots instance.  This means that if we render a gallery of plots all of them will display redundant warning messages.

We should discuss a design we want but in any case, we should pick a model that is consistent with the displays.

a) Local logger for each plot and displays the plot warning independent for each plot
or b) Global logger and displays the warning as a separate component.

kanitw:
Should we close this one as we now have #597?

PR #422, opened 2017-07-11
Filter for quantitative type

starry97:
1. added filter ui for quantitative type
2. added `FILTER_CLEAR` and `FILTER_MODIFY_BOTH_BOUNDS` action
part of #373


src/components/encoding-pane/range-filter-shelf.scss
```diff
@@ -0,0 +1,324 @@
...
+}
+
+.range-filter-pane {
+  width: 80%;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+
+:global {
```
starry97: For some reasons the styles in package rc-slider were not working. I had to copy paste the css file it uses.

starry97:
screen shot 2017-07-10 at 6 12 27 pm
User can click on min/max to edit. This is a pretty bad design now, since it doesn't have any edit affordance.

starry97:
updated UI 
screen shot 2017-07-13 at 1 31 59 pm

src/actions/dataset.ts
```diff
@@ -63,6 +61,8 @@ export function datasetLoad(name: string, dataset: Data): DatasetLoad {
         .then(response => response.json()) // TODO: handle error
         .then(data => fetchCompassQLBuildSchema(data, config)) // TODO: handle error
         .then(schema => {
+          // Clear filters
+          dispatch({ type: FILTER_CLEAR});
```
kanitw: Why SHELF_CLEAR happens after URL_RECEIVE but FILTER_CLEAR happen before?

src/actions/dataset.ts
```diff
@@ -75,6 +75,8 @@ export function datasetLoad(name: string, dataset: Data): DatasetLoad {
     } else if (isInlineData(dataset)) {
       return fetchCompassQLBuildSchema(dataset.values, config) // TODO: handle error
         .then(schema => {
+          // clear filters
+          dispatch({ type: FILTER_CLEAR });
```
kanitw: Why SHELF_CLEAR happens after INLINE_RECEIVE but FILTER_CLEAR happen before?

src/components/encoding-pane/filter-shelf.tsx
```diff
@@ -60,24 +61,25 @@ class FilterShelfBase extends React.Component<FilterShelfProps, {}> {
   private renderFilterShelf(filter: RangeFilter | OneOfFilter, index: number) {
     const {fieldDefs, schema} = this.props;
     const fieldIndex = schema.fieldNames().indexOf(filter.field);
-    const domain = schema.domain(fieldDefs[fieldIndex] as FieldQuery);
+    const fieldDef = fieldDefs[fieldIndex];
+    const domain = schema.domain(fieldDef as FieldQuery);
     return (
-      <div styleName='filter-shelf' key={filter.field}>
+      <div styleName='filter-shelf' key={index}>
```
kanitw: why using `index`  as the key? 

(If you have good reason, add comment to explain?)

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -1,46 +1,123 @@
...
+
     return (
-      <div>
-        {domain}
+      <div styleName='range-filter-pane'>
+        <div>
+          <div styleName='bound'>
+            min: <a onClick={this.focusInput.bind(this, `${filter.field}min`)}><i className="fa fa-pencil"/></a>
+            <input
+              id={`${filter.field}min`}
```
kanitw: _min?

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -1,46 +1,123 @@
...
+              type='text'
+              value={currMin.toString()}
+              onChange={this.filterModifyMinBound.bind(this)}
+            />
+          </div>
+          <div styleName='bound'>
+            max: <a onClick={this.focusInput.bind(this, `${filter.field}max`)}><i className="fa fa-pencil"/></a>
+            <input
+              id={`${filter.field}max`}
```
kanitw: _max?

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -1,46 +1,123 @@
...
+              type='text'
```
kanitw: For quantitative, why not type = number? 
For temporal, I guess clicking this should open some date time picker UI (which can be a separate PR -- but please file an issue).

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -1,46 +1,123 @@
...
+              value={currMax.toString()}
+              onChange={this.filterModifyMaxBound.bind(this)}
+            />
+          </div>
+        </div>
+        <Range
+          allowCross={false}
```
kanitw: What is `allowCross`?

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -1,46 +1,123 @@
...
+      return value;
+    }
+
     return (
-      <div>
-        {domain}
+      <div styleName='range-filter-pane'>
+        <div>
+          <div styleName='bound'>
+            min: <a onClick={this.focusInput.bind(this, `${filter.field}min`)}><i className="fa fa-pencil"/></a>
```
kanitw: We normally do `.bind(this)` in constructor, which mean binding only happen once.  
Doing it in event handler like this will cause .bind() to run every single time. 

Same for the rest of this file.

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -1,46 +1,123 @@
...
+              onChange={this.filterModifyMaxBound.bind(this)}
+            />
+          </div>
+        </div>
+        <Range
+          allowCross={false}
+          defaultValue={[Number(currMin), Number(currMax)]}
+          min={lowerBound}
+          max={upperBound}
+          onAfterChange={this.filterModifyBothBounds.bind(this)}
```
kanitw: Could you please explain why do we need to "modify both bound"?

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -1,46 +1,123 @@
...
+    } else {
+      currMin = range[0];
+      currMax = range[1];
+    }
+    const createSliderWithTooltip = Slider.createSliderWithTooltip;
+    const Range = createSliderWithTooltip(Slider.Range);
+
+    function formatTime(value: number) {
+      if (type === ExpandedType.TEMPORAL) {
+        return new Date(value).toString();
```
kanitw: We could re-use the time formatter that Vega derives from D3.  
(Please file an issue so we can handle this in a separate PR.)

kanitw:
Another question -- if we already have filter for field A in the filter list, if we drag field a to the filter one more time -- what would happen?  I guess we should only allow a field to appear only once.  
(However, later once we support timeUnit, we might want to allow same temporal fields with different timeUnits.)

src/components/encoding-pane/filter-shelf.tsx
starry97: If we want to support multiple filters for one field, we should use index other than field to avoid conflict.

src/actions/dataset.ts
starry97: @kanitw I think they should all happen before `DATESET_URL_RECEIVE`, since we should clear the state before a new dataset is loaded. `SHELF_CLEAR` was written before, so I didn't change it. Is there a particular reason it should happen after DATASET_URL_RECEIVE?

kanitw: I'm not sure about that but they should be consistent.  Maybe move shelf clear to before for consistency?

src/components/encoding-pane/filter-shelf.tsx
kanitw: Should we really support multiple filters for one `{field, timeUnit}`?   

(I'm not sure -- maybe look at how Tableau does this?)

src/components/encoding-pane/range-filter-shelf.tsx
starry97: whether allow slider handles to cross, i.e. whether the min handle can cross the max handle.

src/components/encoding-pane/range-filter-shelf.tsx
starry97: Since we cannot detect whether the min handle is moved or the max handle is moved, we need to call both filterModifyMinBound and filterModifyMaxBound. I think it'd be better to dispatch one action instead of two.

starry97:
> if we already have filter for field A in the filter list, if we drag field a to the filter one more time -- what would happen? I guess we should only allow a field to appear only once.
(However, later once we support timeUnit, we might want to allow same temporal fields with different timeUnits.)
@kanitw it will log an error in the console, and the second filter will not be added.

kanitw:
> @kanitw it will log an error in the console, and the second filter will not be added.

Hmm, this will confuse end-user since they do not know what happen.

PR #424, opened 2017-07-14
fixed previewing shelf cannot be removed

starry97:
bug fix

Previous bug: cannot remove specifying spec shelf. 
preview-bug


kanitw:
Nice!

PR #425, opened 2017-07-17
click outside to close popup

### LINKED ISSUE #285: Support clicking outside function popup to disable
(Right now you can only click caret to disable)

starry97:
1. Click outside to close popup. Fixed #285
2. Refactor field wrapper into a separate class

### DIFF
--- a/src/components/encoding-pane/encoding-shelf.tsx
+++ b/src/components/encoding-pane/encoding-shelf.tsx
@@ -2,17 +2,15 @@ import {isWildcard} from 'compassql/build/src/wildcard';
 import * as React from 'react';
 import * as CSSModules from 'react-css-modules';
 import {ConnectDropTarget, DropTarget, DropTargetCollector, DropTargetSpec} from 'react-dnd';
-import * as TetherComponent from 'react-tether';
-
-import * as styles from './encoding-shelf.scss';
-
 import {ActionHandler} from '../../actions/index';
 import {
   SHELF_FIELD_ADD, SHELF_FIELD_MOVE, SHELF_FIELD_REMOVE, SHELF_FUNCTION_CHANGE, ShelfEncodingAction
 } from '../../actions/shelf';
 import {DraggableType, FieldParentType} from '../../constants';
-import {ShelfFieldDef, ShelfFunction, ShelfId} from '../../models';
+import {ShelfFieldDef, ShelfId} from '../../models';
+import {ShelfFunction} from '../../models/shelf/encoding';
 import {DraggedFieldIdentifier, Field} from '../field/index';
+import * as styles from './encoding-shelf.scss';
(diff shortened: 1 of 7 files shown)


kanitw:
This is awesome progress!

Two high-level comments:

1. Let's rebase and make sure the `FieldWrapper` class is more general and works with the type changer, which is recently merged.  The trick is to pass FunctionPicker as [children](https://facebook.github.io/react/docs/composition-vs-inheritance.html) of `FieldWrapper` instead of including `<FunctionPicker> in the component definition. 

2. I wonder if all of these functionality should just be a part of `<Field/>`
If you disagree and think it should be separate, maybe it should be named `FieldWithPopup`?

src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -65,36 +59,17 @@ class EncodingShelfBase extends React.PureComponent<EncodingShelfProps, Encoding
...
-
-          { this.state.functionPopupOpen &&
-            <FunctionPicker fieldDef={fieldDef} onFunctionChange={this.onFunctionChange}/>
-          }
-        </TetherComponent>
-      </div>
+      <FieldWrapper
+        id={id}
+        fieldDef={fieldDef}
+        functionPopupOpen={this.state.functionPopupOpen}
```
kanitw: I think `functionPopupOpen`, `toggleFunctionPopup` and `closeFunctionPopup` should be maintained internally inside `FieldWrapper`?  

(Why should `EncodingShelf` maintain the state for its internal component?)

src/components/field/index.tsx
```diff
@@ -44,21 +41,54 @@ export interface FieldPropsBase {
...
-class FieldBase extends React.PureComponent<FieldProps, {}> {
+export interface FieldState {
+  popupIsOpened?: boolean;
+}
+
+class FieldBase extends React.PureComponent<FieldProps, FieldState> {
+  private field: HTMLElement;
+  private popup: HTMLElement;
+  private fieldRefHandler = {
+    field: (ref: any) => this.field = ref
```
kanitw: private fieldRefHandler(ref: any) {
  this.field = ref;
}

kanitw: (ref: any) => this.field = ref

```
function (ref) {
   return this.field = ref;
}
```

```
function (ref) {
  this.field = ref;
}
```

src/components/field/index.tsx
```diff
@@ -56,10 +56,14 @@ class FieldBase extends React.PureComponent<FieldProps, FieldState> {
   private field: HTMLElement;
   private popup: HTMLElement;
   private fieldRefHandler = {
-    field: (ref: any) => this.field = ref
+    field: (ref: any) => {
```
kanitw: remove the nested field too, same for popup.

PR #426, opened 2017-07-17
[Need to revise] encoding preview trigger plot re-rendering

### LINKED ISSUE #308: Encoding Preview Trigger Plot Re-rendering
Try adding Cylinders and Quantitative Fields Wildcards

starry97:
fixed #308

Note: If I understood this issue correctly, the problem is when hovering over an encoding preview, the plot moves slightly. This is fixed in this PR.


kanitw:
I think this should be fixed with https://github.com/vega/voyager/issues/431 instead.   
Correct me if I'm wrong.

PR #439, opened 2017-07-20
Add api to load a vega-lite spec into voyager.

### LINKED ISSUE #377: API for loading visual encoding
For example, if the electron app wanna support loading .vl.json file, we need to be able to update the shelf

tafsiri:
Add api to load a vega-lite spec into voyager. Note that this will clear any previously set data. Closes #377

Ping @kanitw. Finally got this going, tests took longer to debug/make run correctly than I expected.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -1,5 +1,11 @@
 import './app.scss';
 
+import * as Ajv from 'ajv';
+import * as draft4Schemas from 'ajv/lib/refs/json-schema-draft-04.json';
+import {fromSpec} from 'compassql/build/src/query/spec';
+import { FacetedCompositeUnitSpec, isUnitSpec, TopLevel } from 'vega-lite/build/src/spec';
+import * as vlSchema from 'vega-lite/build/vega-lite-schema.json';
+
 import * as React from 'react';
 import {DragDropContext} from 'react-dnd';
 import HTML5Backend from 'react-dnd-html5-backend';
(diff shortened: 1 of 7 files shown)


src/components/app.tsx
```diff
@@ -89,6 +103,60 @@ class AppBase extends React.PureComponent<Props, {}> {
...
+      allErrors: true,
+      extendRefs: 'fail'
+    });
+    ajv.addMetaSchema(draft4Schemas, 'http://json-schema.org/draft-04/schema#');
+
+    const validateVl = ajv.compile(vlSchema);
+    const valid = validateVl(spec);
+
+    if (!valid) {
+      throw new Error("Invalid Schema:" + validateVl.errors.toString());
```
kanitw: Invalid spec?  (Schema is not invalid, but the spec does not conform to the schema?)

src/components/app.tsx
```diff
@@ -89,6 +103,60 @@ class AppBase extends React.PureComponent<Props, {}> {
...
+
+    const validSpec: TopLevel<FacetedCompositeUnitSpec> = spec as TopLevel<FacetedCompositeUnitSpec>;
+
+
+    if (!isUnitSpec(validSpec)) {
+      throw new Error("Voyager does not support layered or multi-view vega-lite specs");
+    }
+
+    if (validSpec.data) {
+      const setDataAction = this.setData(validSpec.data) as Promise<any>;
```
kanitw: not sure why do you have to cast `as Promise<any>` here.   

I wonder if we should just void casting by just doing
```
this.setData(validSpec.data)
  .then(...
```

src/components/app.tsx
tafsiri: typo. got it.

src/components/app.tsx
tafsiri: good catch

tafsiri:
@kanitw made changes according to your suggestions.

kanitw:
Thanks!

PR #441, opened 2017-07-20
Load header images with web pack loader.

tafsiri:
This is a fix for https://github.com/vega/voyager-electron/issues/10, using the webpack will allow the files to be copied over during the build.

ping @kanitw @starry97


kanitw:
Thanks!

PR #445, opened 2017-07-21
do not re-create spec

starry97:


src/selectors/index.ts
```diff
@@ -31,6 +46,17 @@ export const getFilters = (state: State) => state.present.shelf.spec.filters;
 export const getShelf = (state: State) => state.present.shelf;
 export const getSchema = (state: State) => state.present.dataset.schema;
 export const getMainResult = (state: State) => state.present.result.main.modelGroup;
+export const getSpec = (state: State) => {
```
kanitw: 1) You should use `createSelector` to correctly chain the dependencies.
(Maybe this is why it does not solve the problem yet.) 

2) rename this `getMainSpec` and make it depends on `query` as well.

src/components/view-pane/index.tsx
```diff
@@ -1,32 +1,31 @@
...
-  filters: Array<RangeFilter | OneOfFilter>;
   mainResult: SpecQueryGroup<PlotObject>;
   bookmark: Bookmark;
+  spec: FacetedCompositeUnitSpec;
 }
 
 class ViewPaneBase extends React.PureComponent<ViewPaneProps, {}> {
   public render() {
-    const {bookmark, data, handleAction, filters, query, mainResult} = this.props;
+    const {bookmark, handleAction, query, mainResult, spec} = this.props;
```
kanitw: `query` and `mainResult` should no longer be a prop of this component. 

We should instead have `mainSpec` and `mainPlots` and at most one of them can have value at a particular time.

The selector for `mainSpec` and `mainPlots` should depend on `query` and the full `mainResult`  -- not just `result.main.modelGroup`.  Thus, we can read `result.main.isLoading`.

If `result.main.isLoading` is`true`, both `mainSpec` and `mainPlots` should be `undefined`.

src/selectors/index.ts
kanitw: See https://github.com/vega/voyager/pull/445#discussion_r128854359 too -- both comments are relevant :)

src/components/view-pane/index.tsx
```diff
@@ -1,65 +1,53 @@
...
 
+
+
 export interface ViewPaneProps extends ActionHandler<ShelfAction> {
-  data: Data;
-  query: Query;
-  filters: Array<RangeFilter | OneOfFilter>;
-  mainResult: SpecQueryGroup<PlotObject>;
+  mainResult: {
+    mainSpec: FacetedCompositeUnitSpec,
```
kanitw: spec, plots  (it's already clear that this is for main with `mainResult`

src/components/view-pane/index.tsx
```diff
@@ -68,12 +56,9 @@ class ViewPaneBase extends React.PureComponent<ViewPaneProps, {}> {
 export const ViewPane = connect(
   (state: State) => {
     return {
-      data: selectData(state),
-      query: selectQuery(state),
-      filters: selectFilters(state),
       // FIXME: refactor the flow for this part (we should support asynchrounous request for this too)
       mainResult: selectMainResult(state),
-      bookmark: selectBookmark(state)
+      bookmark: selectBookmark(state),
```
kanitw: why extra comma?

src/selectors/index.ts
```diff
@@ -67,3 +81,46 @@ export const selectSchemaFieldDefs = createSelector(
     });
   }
 );
+
+const selectResult = (state: State) => state.present.result.main;
+
+const selectMainSpec = createSelector(
```
kanitw: Why do we need a separate selector here? 

We can simply use

```
{
  data: data,
  transform: getTransforms(filters),
  ...getTopSpecQueryItem(result.modelGroup).spec
}
```

in `selectMainResult`

src/selectors/index.ts
```diff
@@ -67,3 +81,46 @@ export const selectSchemaFieldDefs = createSelector(
...
+    }
+    return {
+      data: data,
+      transform: getTransforms(filters),
+      ...getTopSpecQueryItem(result.modelGroup).spec
+    };
+  }
+);
+
+export const selectMainResult = createSelector(
```
kanitw: name this `selectMainResultForView` (so `selectResult` can be `selectMainResult` like before?)

PR #454, opened 2017-07-24
[WIP] All Filter Combined

kanitw:
Final check list

- [x] use `PureComponent` for all of our components are we guarantee that our props are immutable.


src/components/encoding-pane/filter-shelf.tsx
```diff
@@ -0,0 +1,127 @@
...
+export interface FilterShelfPropsBase {
+  filters: Array<RangeFilter | OneOfFilter>;
+  fieldDefs: ShelfFieldDef[];
+  schema: Schema;
+  handleAction?: (action: FilterAction) => void;
+}
+
+interface FilterShelfProps extends FilterShelfDropTargetProps, FilterShelfPropsBase {};
+
+class FilterShelfBase extends React.Component<FilterShelfProps, {}> {
```
kanitw: @starry97 We should use `PureComponent` for all of our components are we guarantee that our props are immutable.

PR #455, opened 2017-07-25
time unit filter

starry97:
screen shot 2017-07-25 at 10 35 50 am
screen shot 2017-07-25 at 10 36 27 am


src/components/encoding-pane/filter-shelf.scss
```diff
@@ -22,9 +22,21 @@ $radius: 3px;
   }
 }
 
-.header {
```
kanitw: I wonder why we have to remove header here.  (Why is it relevant to adding a timeUnit filter?)

src/components/encoding-pane/filter-shelf.tsx
```diff
@@ -31,9 +37,20 @@ export interface FilterShelfPropsBase {
   handleAction?: (action: FilterAction) => void;
 }
 
+export interface FilterShelfState {
+  timeUnit: boolean;
```
kanitw: maybe `hasTimeUnit`?   (Normally `timeUnit` should have a type `TimeUnit`)

kanitw: Maybe you don't need state here?

src/components/encoding-pane/filter-shelf.scss
starry97: the style is no longer used because we use field for the header

starry97: Done

src/components/encoding-pane/filter-shelf.tsx
```diff
@@ -58,31 +68,128 @@ class FilterShelfBase extends React.Component<FilterShelfProps, {}> {
...
+          range: domain
+        }
+      });
+      return;
+    } else if (timeUnit === TimeUnit.MONTH || timeUnit === TimeUnit.DAY) {
+      handleAction({
+        type: FILTER_MODIFY_ONE_OF,
+        payload: {
+          index,
+          oneOf: getRange(domain, timeUnit)
```
kanitw: `oneOfFilter` but call `getRange`?   

I wonder if you should split between getDefaultRange (`[min, max]`) and getDefaultList (full array of all possible values).

src/components/encoding-pane/filter-shelf.tsx
```diff
@@ -58,31 +68,128 @@ class FilterShelfBase extends React.Component<FilterShelfProps, {}> {
...
+    }
+
+    handleAction({
+      type: FILTER_MODIFY_TIME_UNIT,
+      payload: {
+        index,
+        timeUnit
+      }
+    });
+    if (!timeUnit) {
```
kanitw: I'm wondering if some of these logic down here should be handled as a part of `FILTER_MODIFY_TIME_UNIT`.  

Doing so would at least make it easier to write unit tests, but the trade off is that filterReducer will probably need to take schema as input just like we do for shelf.
(I think that's fine.  Otherwise, we should similarly get rid of `schema` from `shelfReducer` too.) 

I'm not sure though.  If you disagree we can discuss

src/components/encoding-pane/filter-shelf.tsx
```diff
@@ -93,6 +200,36 @@ class FilterShelfBase extends React.Component<FilterShelfProps, {}> {
       </span>
     );
   }
+
+  private renderTimeUnitChanger(filter: RangeFilter | OneOfFilter, index: number, domain: number[]) {
```
kanitw: Why don't we re-use `FunctionPicker` so we can have less code and ensure consistent style?

src/components/encoding-pane/filter-shelf.tsx
```diff
@@ -58,31 +68,128 @@ class FilterShelfBase extends React.Component<FilterShelfProps, {}> {
...
-          <span>{filter.field}</span>
-          <a onClick={this.filterRemove.bind(this, index)}>
-            <i className='fa fa-times'/>
-          </a>
-        </div>
-        {this.renderFilter(filter, index, domain, fieldDef.type)}
-      </div>
+    let domain = schema.domain(fieldDef as FieldQuery);
+    const popupComponent = (
+      fieldDef.type === ExpandedType.TEMPORAL ?
```
kanitw: you can do  
```
fieldDef.type === ExpandedType.TEMPORAL && 
this.renderTimeUnitChanger(filter, index, domain)
```

Probably nicer than using ternary operator

src/components/encoding-pane/filter-shelf.tsx
```diff
@@ -58,31 +68,128 @@ class FilterShelfBase extends React.Component<FilterShelfProps, {}> {
...
-    const {handleAction} = this.props;
+    let filterComponent;
     if (isRangeFilter(filter)) {
-      return <RangeFilterShelf domain={domain} index={index} filter={filter} handleAction={handleAction} type={type}/>;
+      const timeUnit = filter.timeUnit;
+      if (timeUnit) {
+        domain = getRange(domain, timeUnit);
+        if (timeUnit === TimeUnit.MONTH || timeUnit === TimeUnit.DAY) {
+          filterComponent = (
+            <OneOfFilterShelf
```
kanitw: Why is `OneOfFilterShelf` the filter component  inside `if(isRangeFilter(filter)` ?

src/components/encoding-pane/filter-shelf.tsx
```diff
@@ -6,13 +6,19 @@ import * as React from 'react';
 import * as CSSModules from 'react-css-modules';
```
kanitw: Let's rename this file and component to filter-pane.

Since you have shelf as a shelf for single filter, calling the grouping of multiple shelves a shelf too is a bit confusing :)

src/components/field/index.tsx
```diff
@@ -222,9 +222,11 @@ class FieldBase extends React.PureComponent<FieldProps, FieldState> {
   }
 
   private closePopup() {
-    this.setState ({
-      popupIsOpened: false
-    });
+    if (this.state) {
+      this.setState ({
```
kanitw: `setState(`   <-- Please remove before `(`.

src/components/field/field.scss
```diff
@@ -1,3 +1,5 @@
+@import '../encoding-pane/filter-shelf.scss';
```
kanitw: Why do we need to do an import here?  Can't we make the relevant component self-contained?

kanitw: Actually looking at `encoding-pane/filter-shelf.scss` I think we probably won't need this as you use this for timeUnitChanger as we probably should use FunctionPicker anyway. :)

src/reducers/shelf/filter.test.ts
```diff
@@ -160,4 +161,20 @@ describe('reducers/shelf/filter', () => {
       expect(filter).toEqual(rangeFilter);
     });
   });
+
+  describe(FILTER_MODIFY_TIME_UNIT, () => {
+    it('should add an time unit to the current filter', () => {
```
kanitw: a time unit :)

src/reducers/shelf/filter.ts
```diff
@@ -114,6 +115,20 @@ export function filterReducer(shelfSpec: Readonly<ShelfUnitSpec> = DEFAULT_SHELF
...
     }
+    case FILTER_MODIFY_TIME_UNIT: {
+      const {index, timeUnit} = action.payload;
+      const modifier = (filter: RangeFilter) => {
+        return {
+          ...filter,
+          timeUnit
+        };
+      };
+      const filters = modifyItemInArray(shelfSpec.filters, index, modifier);
```
kanitw: I personally prefer if we just do
```
filters: modifyItemInArray(shelfSpec.filters, index, modifier)
```

below as it is less code.

src/reducers/shelf/filter.ts
```diff
@@ -114,6 +115,20 @@ export function filterReducer(shelfSpec: Readonly<ShelfUnitSpec> = DEFAULT_SHELF
         filters
       };
     }
+    case FILTER_MODIFY_TIME_UNIT: {
+      const {index, timeUnit} = action.payload;
+      const modifier = (filter: RangeFilter) => {
```
kanitw: I would either put this inline below or name the `modifier` "`modifyTimeUnit`"

src/components/encoding-pane/function-picker.tsx
```diff
@@ -19,17 +25,29 @@ export class FunctionPickerBase extends React.PureComponent<FunctionPickerProps,
...
   }
   public render() {
-    const {fieldDef} = this.props;
-
-    const fn = fieldDef.aggregate || fieldDef.timeUnit || (fieldDef.bin && 'bin' ) || undefined;
+    const {fieldDef, filter, index} = this.props;
+    let fn: any, name: string;
+    if (filter) {
+      fn = filter.timeUnit;
+      name = index.toString();
```
kanitw: Maybe "filter_" + index?  -- just `index` isn't very descriptive.

src/components/encoding-pane/function-picker.tsx
```diff
@@ -19,17 +25,29 @@ export class FunctionPickerBase extends React.PureComponent<FunctionPickerProps,
...
+    } else {
+      fn = fieldDef.aggregate || fieldDef.timeUnit || (fieldDef.bin && 'bin' ) || undefined;
+      name = JSON.stringify(fieldDef);
```
kanitw: I assume that we need `name` otherwise it is buggy.  

If so, if filter uses index, why don't we use shelfId here? 
I think maybe instead of making both `index` and `shelfId` a parameter, why don't we make name a parameter of FunctionPicker?

src/components/encoding-pane/function-picker.tsx
```diff
@@ -2,13 +2,19 @@ import {ExpandedType} from 'compassql/build/src/query/expandedtype';
...
+import {OneOfFilter, RangeFilter} from 'vega-lite/build/src/filter';
+import {TimeUnit} from 'vega-lite/build/src/timeunit';
 import {ShelfFieldDef, ShelfFunction} from '../../models/shelf';
 import * as styles from './function-picker.scss';
 
 export interface FunctionPickerProps {
   fieldDef: ShelfFieldDef;
 
-  onFunctionChange: (fn: ShelfFunction) => void;
+  filter?: RangeFilter | OneOfFilter;
```
kanitw: Instead of adding `filter` as another possible params, I wonder if we should just change `fieldDef: ShelfFieldDef`

to 
````
fieldDefParts: {
   aggregate: ...,
   bin: ...,
   timeUnit: ...,
   field: ...
} 
```

which will be reuseable for both ShelfFieldDef and Filter.

src/components/encoding-pane/function-picker.tsx
```diff
@@ -43,7 +61,12 @@ export class FunctionPickerBase extends React.PureComponent<FunctionPickerProps,
     }
   }
   private onFunctionChange(event: any) {
-    this.props.onFunctionChange(event.target.value as ShelfFunction);
+    const {index} = this.props;
+    if (index !== undefined) {
+      this.props.onFunctionChange(event.target.value, index);
```
kanitw: A better design of a reusable component would minimize unnecessary parameterization.  
Thus, I wonder if there is a way to structure this without requiring `index`.  

(At least, the any encoding shelf, which similarly is a list, also does not require index in `onFunctionChange`.)

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -199,12 +197,12 @@ export class RangeFilterShelfBase extends React.Component<RangeFilterShelfProps,
   }
 
   // TODO: https://github.com/vega/voyager/issues/443: use the time formatter Vega derives from D3
-  private formatTime = (value: number) => {
+  private formatTime = (value: number) : React.ReactText => {
```
kanitw: `(value: number): React.ReactText`  (no space after `)`)

Also wonder why it has to be `React.ReactText`?  What's wrong with just being a string?

src/reducers/shelf/filter.ts
```diff
@@ -145,3 +186,54 @@ export function getFilter(fieldDef: ShelfFieldDef, domain: any[]): RangeFilter |
...
+      throw new Error ('Invalid range time unit ' + timeUnit);
+  }
+}
+
+export function getDefaultList(timeUnit: TimeUnit): number[] {
+  switch (timeUnit) {
+    case TimeUnit.MONTH:
+      return [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12];
+    case TimeUnit.DAY:
+      return [1, 2, 3, 4, 5, 6, 7];
```
kanitw: Do we have an issue filed for changing label for MONTH and DAY? 

(Month is not that confusing, but I have no idea what 1 means for DAY (Sunday / Monday?)

src/reducers/shelf/filter.ts
```diff
@@ -94,24 +98,61 @@ export function filterReducer(shelfSpec: Readonly<ShelfUnitSpec> = DEFAULT_SHELF
...
+        type: 'temporal'
+        // TODO: remove channel and type after it's fixed in compassql
+      });
+      let modifyTimeUnit;
+      if (!timeUnit) {
+        modifyTimeUnit = (filter: RangeFilter) => {
+          return {
+            field: filter.field,
+            timeUnit,
+            range: domain as number[] | DateTime[]
```
kanitw: Why do we have to do casting here?

src/components/encoding-pane/filter-pane.tsx
```diff
@@ -0,0 +1,226 @@
...
+  connectDropTarget: ConnectDropTarget;
+
+  isOver: boolean;
+
+  item: Object;
+}
+
+export interface FilterPanePropsBase {
+  filters: Array<RangeFilter | OneOfFilter>;
+  fieldDefs: ShelfFieldDef[];
```
kanitw: Why does filter pane requires an array of FieldDefs?

src/components/encoding-pane/filter-pane.tsx
```diff
@@ -0,0 +1,226 @@
...
+  schema: Schema;
```
kanitw: Do we still need `schema`?

src/components/encoding-pane/filter-pane.tsx
```diff
@@ -0,0 +1,226 @@
...
+        timeUnit: timeUnit as TimeUnit
+      }
+    });
+  }
+
+  private renderFilterShelf(filter: RangeFilter | OneOfFilter, index: number) {
+    const {fieldDefs, schema, handleAction} = this.props;
+    const fieldIndex = schema.fieldNames().indexOf(filter.field);
+    const fieldDef = fieldDefs[fieldIndex];
+    let domain = schema.domain(fieldDef as FieldQuery);
```
kanitw: Seeing this line, I assume this PR is not ready?  (I remember helping you a bit on this this morning.)

src/components/encoding-pane/filter-pane.tsx
starry97: Because I need the fieldDef field and type to construct an filedQuery. This can be removed after the domain function in compassql is modified.

src/components/encoding-pane/filter-pane.tsx
starry97: Yes, because I need to call schema.domain()

src/components/encoding-pane/filter-pane.tsx
kanitw: Wait, but yesterday morning I have already shown you in person that you can just construct a new fieldQuery with a right field in it (and just use `channel: SHORT_WILDCARD`).

starry97: Oh wait, I still need to use fieldDef to check whether to render the function picker caret (so I need to check whether its type is temporal).

src/reducers/shelf/filter.ts
starry97: Is the issue in vega-lite because it throws errors when January etc. is used?

src/components/encoding-pane/filter-pane.tsx
kanitw: Note from personal discussion: we can read the type from the `schema`?

src/components/encoding-pane/function-picker.tsx
starry97: Yea, I could extract single filter shelf as one file, so that I can read index as a prop, just like the encoding shelf. @kanitw Do you think this is worth doing?

src/components/encoding-pane/filter-pane.tsx
```diff
@@ -0,0 +1,226 @@
...
+            />
+          );
+        } else {
+          // time unit range filter
+          domain = getDefaultRange(domain, timeUnit);
+          let type;
+          if (timeUnit === TimeUnit.YEARMONTHDATE) {
+            type = ExpandedType.TEMPORAL;
+          } else {
+            type = ExpandedType.QUANTITATIVE;
```
starry97: I set the type to TEMPORAL so that it'd render date time pickers. Set it to quantitative just renders inputs.

src/components/encoding-pane/filter-pane.tsx
kanitw: Okie :)

src/components/encoding-pane/filter-pane.tsx
kanitw: But why do some temporal data with some timeUnit should use input (not date pickers)?

src/components/encoding-pane/filter-pane.scss
```diff
@@ -0,0 +1,49 @@
...
+.filter-shelf {
+  background-color: white;
+  border-radius: $radius;
+  border: 1px solid #ddd;
+  margin-bottom: 4px;
+  max-height: 150px;
+  overflow: scroll;
+
+  :global .fa {
+    &.fa-times {
```
kanitw: Isn't this already declared somewhere else?

src/components/encoding-pane/filter-pane.scss
```diff
@@ -0,0 +1,49 @@
...
+
+      &:hover {
+        color: #fff;
+        background-color: #999;
+      }
+    }
+  }
+}
+
+.placeholder {
```
kanitw: We should design a way to share common styles between encodings and filter shelves instead of  just copy and paste style for the similar things -- otherwise it is hard to maintain consistency.

src/components/encoding-pane/filter-pane.tsx
```diff
@@ -0,0 +1,205 @@
...
+      <FunctionPicker
+        fieldDefParts={{
+          timeUnit: filter.timeUnit,
+          type: ExpandedType.TEMPORAL
+        }}
+        onFunctionChange={onFunctionChange}
+      /> ;
+    let filterComponent;
+    let type: ExpandedType = fieldDef.type === ExpandedType.TEMPORAL ?
+     ExpandedType.TEMPORAL : ExpandedType.QUANTITATIVE;
```
kanitw: use two space for tab, not one :)

src/components/encoding-pane/filter-pane.tsx
```diff
@@ -0,0 +1,205 @@
...
+    const timeUnit = filter.timeUnit;
+    if (isRangeFilter(filter)) {
+      if (timeUnit) {
+        domain = getDefaultRange(domain, timeUnit);
+        if (timeUnit !== TimeUnit.YEARMONTHDATE) {
+          type = ExpandedType.QUANTITATIVE;
+          // set the type to quantitative so that it will render
+          // inputs instead of date time changers.
```
kanitw: // normal number input instead of ... ?

src/components/field/index.tsx
```diff
@@ -115,7 +113,7 @@ class FieldBase extends React.PureComponent<FieldProps, FieldState> {
       </span>
     );
     // Wrap with connect dragSource if it is injected
-    if (!popupComponent) {
+    if (!popupComponent || !this.state) {
```
kanitw: Would `!this.state` ever be triggered?

src/reducers/shelf/filter.ts
kanitw: Good point, feel free to file one in Vega-Lite.

src/components/encoding-pane/filter-pane.scss
starry97: Should we have one global .fa class as all font awesome icons we use have the same style?

kanitw: Yeah, I think the `app.scss` which already the file for all `:global` stuff would be an appropriate place :)

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -4,84 +4,100 @@ import * as React from 'react';
...
+    lowerBound = Math.floor(Number(domain[0]));
+    upperBound = Math.ceil(Number(domain[1]));
+    if (isDateTime(domain[0])) {
+      lowerBound = Math.floor(convertToTimestamp(domain[0] as DateTime));
+    }
+    if (isDateTime(domain[1])) {
+      upperBound = Math.ceil(convertToTimestamp(domain[1] as DateTime));
+    }
+    let step;
+    if (filter.timeUnit === TimeUnit.YEARMONTHDATE) {
```
kanitw: What about other timeUnit?  

Let's extract a method for any timeUnit specific rules and we can define step for all supported timeUnit and throw error for ones that we have not supported yet?  (This way when we want to add more, we can easily know we need to add it here.  

(Same for other things that you do specifically for timeUnits that we currently support.)

kanitw: Also, one day seems applicable for temporal with no time unit too?

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -4,84 +4,100 @@ import * as React from 'react';
...
+          step={step}
         />
       </div>
     );
   }
 
-  protected filterModifyExtent(range: number[]) {
+  protected filterModifyExtent(range: number[] | DateTime[]) {
+    if (this.props.type === ExpandedType.TEMPORAL) {
+      range = [convertToDateTimeObject(range[0] as number), convertToDateTimeObject(range[1] as number)];
```
kanitw: why `as number` here -- seems like we better refine the type somewhere?

src/components/encoding-pane/range-filter-shelf.tsx
```diff
@@ -158,6 +183,10 @@ export class RangeFilterShelfBase extends React.Component<RangeFilterShelfProps,
       dateTimePickerOpen = this.state.maxDateTimePickerOpen;
       dataTimePickerOpenAction = this.toggleMaxDateTimePicker;
     }
+    let timeFormat = true;
+    if (this.props.filter.timeUnit === TimeUnit.YEARMONTHDATE) {
```
kanitw: What about other timeUnit?  

Let's extract a method for any timeUnit specific rules and we can define step for all supported timeUnit and throw error for ones that we have not supported yet?  (This way when we want to add more, we can easily know we need to add it here.  

(Same for other things that you do specifically for timeUnits that we currently support.)

src/components/field/index.tsx
```diff
@@ -277,7 +275,10 @@ export interface DraggedFieldIdentifier {
 const fieldSource: DragSourceSpec<FieldProps> = {
   beginDrag(props): DraggedFieldIdentifier {
     const {fieldDef, parentId, schema} = props;
-    const filter = getFilter(fieldDef, schema.domain(fieldDef as FieldQuery));
+    if (isWildcard(fieldDef.field)) {
```
kanitw: Can you reply to explain what is this logic for and also add comment accordingly?

src/components/encoding-pane/range-filter-shelf.tsx
starry97: > Also, one day seems applicable for temporal with no time unit too?

@kanitw The default temporal filter is without time unit.

PR #461, opened 2017-07-25
[WIP] Add Wildcard Function actions and reducers

felixcodes:
#304 - 
Completed:
- `SHELF_FUNCTION_ENABLE_WILDCARD` action/reducer
- `SHELF_FUNCTION_DISABLE_WILDCARD` action/reducer
- `SHELF_FUNCTION_ADD_WILDCARD` action/reducer
- `SHELF_FUNCTION_REMOVE_WILDCARD` action/reducer


kanitw:
Is this done? If so rebase and remove WIP :)

src/reducers/shelf/spec.test.ts
```diff
@@ -364,6 +366,483 @@ describe('reducers/shelf/spec', () => {
...
+
+    it('should correctly convert back to a specific function', () => {
+      const shelfSpec = shelfSpecReducer(
+        {
+          ...DEFAULT_SHELF_UNIT_SPEC,
+          encoding: {
+            x: {
+              aggregate: {name: 'aggregate', enum: ['mean', 'median']},
+              bin: '?',
+              timeUnit: {name: 'timeUnit', enum: ['year']},
```
kanitw: remove `timeUnit` from this test? (SInce it is impossible to have them with aggregate/bin)

src/reducers/shelf/spec.test.ts
```diff
@@ -364,6 +366,483 @@ describe('reducers/shelf/spec', () => {
...
+        }
+      });
+    });
+
+    it('should correctly convert back to a specific function when no wildcard functions were selected', () => {
+      const shelfSpec = shelfSpecReducer(
+        {
+          ...DEFAULT_SHELF_UNIT_SPEC,
+          encoding: {
+            x: {field: 'a', type: 'quantitative'}
```
kanitw: Would disable wildcard be even clickable in this state?  

I guess you want something like `aggregate: {enum: []}`? 

This leads me to another question which is what SHELF_FUNCTION_REMOVE_WILDCARD do when there is only one item left.  (Especially if the last one is bin.)

src/actions/shelf.ts
```diff
@@ -46,21 +47,36 @@ export type ShelfFieldMove = ReduxAction<typeof SHELF_FIELD_MOVE, {
...
+export const SHELF_FUNCTION_ADD_WILDCARD = 'SHELF_FUNCTION_ADD_WILDCARD';
+export type ShelfFunctionAddWildcard = ReduxAction<typeof SHELF_FUNCTION_ADD_WILDCARD, {
+  shelfId: ShelfId,
+  fn: ShelfFunction;
+}>;
+
+export const SHELF_FUNCTION_ENABLE_WILDCARD = 'SHELF_FUNCTION_ENABLE_WILDCARD';
+export type ShelfFunctionEnableWildcard = ReduxAction<typeof SHELF_FUNCTION_ENABLE_WILDCARD, {
+  shelfId: ShelfId,
+  fn: ShelfFunction;
```
kanitw: Since SHELF_FUNCTION_DISABLE_WILDCARD does not require `fn`, SHELF_FUNCTION_ENABLE_WILDCARD should not require `fn`.

src/reducers/shelf/spec.test.ts
```diff
@@ -364,6 +366,483 @@ describe('reducers/shelf/spec', () => {
...
+      expect(shelfSpec).toEqual({
+        ...DEFAULT_SHELF_UNIT_SPEC,
+        encoding: {
+          x: {field: 'a', type: 'quantitative'}
+        }
+      });
+    });
+  });
+
+  describe(SHELF_FUNCTION_ENABLE_WILDCARD, () => {
```
kanitw: I would add a case where previously there is no function, and then you enable wildcard.  
How do we distinguish between (1) no function, wildcard disabled and (2) no function, wildcard enabled.

PR #462, opened 2017-07-26
Use store listener to cause side effect instead.

kanitw:
I think this is definitely an improvement over a hacky solution inside `encoding-pane.ts`.  

That said, if this causes bad effect for loading program state or loading a spec, we could consider other solution.  

In any case, I'll fork from from this branch to work on related views.

### DIFF
--- a/src/actions/result.ts
+++ b/src/actions/result.ts
@@ -12,30 +12,36 @@ import {Action} from './index';
 export type ResultAction = ResultRequest | ResultReceive;
 export type ResultAsyncAction = AsyncResultRequest;
 
+export type ResultType = 'main'; // TODO: add 'summary', etc.
+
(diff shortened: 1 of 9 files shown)


vlandham:
ah. but you are getting error: 

> Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the Connect(BookmarkBase) component.

kanitw:
@vlandham  somehow it's not trigger the same problem anymore

vlandham:
great!

kanitw:
Since it is passing, I will merge it first then, but could you help test https://github.com/vega/voyager/issues/466 and if this does not work we should revise our approach.

vlandham:
sounds good.

PR #471, opened 2017-07-26
Refactor queries related stuff

kanitw:
(prepare for related views)

### DIFF
--- a/src/models/index.ts
+++ b/src/models/index.ts
@@ -7,7 +7,7 @@ import {duplicate} from "vega-lite/build/src/util";
 import {Bookmark, DEFAULT_BOOKMARK} from './bookmark';
 import {DEFAULT_VOYAGER_CONFIG, VoyagerConfig} from './config';
 import {Dataset, DEFAULT_DATASET} from './dataset';
-import {DEFAULT_RESULT, ResultIndex} from './result';
+import {DEFAULT_RESULT_INDEX, ResultIndex} from './result';
 import {DEFAULT_SHELF_SPEC, Shelf} from './shelf';
 
 export * from './bookmark';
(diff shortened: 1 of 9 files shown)


src/queries/index.ts
```diff
@@ -0,0 +1,10 @@
+import {Query} from 'compassql/build/src/query/query';
+import {Store} from 'redux';
+import {StateWithHistory} from 'redux-undo';
+import {StateBase} from '../../build/src/models/index';
+import {resultRequest} from '../actions/result';
+
+export function makeQueries(store: Store<StateWithHistory<Readonly<StateBase>>>, query: Query) {
```
vlandham: the verb 'make' is consistent with other areas in the code where we dispatch ? or is this the first time this is happening?

src/models/index.ts
```diff
@@ -37,7 +37,7 @@ export const DEFAULT_STATE: StateBase = {
   config: DEFAULT_VOYAGER_CONFIG,
   dataset: DEFAULT_DATASET,
   shelf: DEFAULT_SHELF_SPEC,
-  result: DEFAULT_RESULT,
+  result: DEFAULT_RESULT_INDEX,
```
starry97: I'm not sure why you changed the name to DEFAULT_RESULT_INDEX. It looks like we are using ResultIndex in other places too, but it seems like a confusing name to me.

kanitw: `state.result` stores an dictionary (index) for different results.    

`result.main` is for the main result, but we will have other results too.

src/queries/index.ts
kanitw: I'll name it `dispatchQueries` then.

kanitw:
I'm making this depends on https://github.com/vega/voyager/pull/476.

PR #475, opened 2017-07-26
Fix why result disappear when a new bookmark is added

kanitw:


felixcodes:
Looks good. Will merge when Travis finishes.

PR #477, opened 2017-07-26
moved copied indicator to be below plot buttons

### LINKED ISSUE #469: click on copy spec shouldn't move the plot

starry97:
fixed #469

### DIFF
--- a/src/components/plot/index.tsx
+++ b/src/components/plot/index.tsx
@@ -30,6 +30,7 @@ export interface PlotProps extends ActionHandler<ShelfAction | BookmarkAction> {
 export interface PlotState {
   hovered: boolean;
   preview: boolean;
+  copiedPopupIsOpened: boolean;
 }
 
 export class PlotBase extends React.PureComponent<PlotProps, any> {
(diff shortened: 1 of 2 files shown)


src/components/plot/plot.scss
```diff
@@ -48,4 +48,11 @@
 
 .copy-button {
   cursor: pointer;
+  position: relative;
+}
+
+.copied {
```
kanitw: If you use Tether anyway, why do we still need `.copied`?

kanitw: That said, I'd say keep the styleName but remove the position/left/margin below.

Instead, let's make the font-size a bit smaller.  It's quite too big currently.

src/components/plot/index.tsx
```diff
@@ -229,7 +235,10 @@ export class PlotBase extends React.PureComponent<PlotProps, any> {
   }
 
   private copied() {
-    const copiedIndicator = document.getElementById('copied');
+    const copiedIndicator = document.getElementById(JSON.stringify(this.props.spec));
```
kanitw: Instead of causing side-effect using `document.getElementById`, 
why don't we do things in react way -- e.g., by making the popup depends on a local state that determines if a popup should be shown.

Issue #385, opened 2017-06-16
Benchmarking


kanitw:
- [ ] Agree on what to benchmark

- [ ] (client mode vs server mode)
- [ ] maximum file size
- [ ] when should we turn related views off

tafsiri:
- How big a data file can be loaded in voyager? How long does it take between selecting some data and the app being ready?

tafsiri:
Benchmarking Ideas

 - [ ] Timing information between dispatched actions.
 - [ ] Time to finish render after some major task
    - [ ] Load a data file. How long until app is ready (all plots are rendered?)
    - [ ] Put something on the shelf (can be done via loading a spec that has inline data)

vlandham:
Clarification from call: We want to focus on render time. Specifically, `plot-list` appears to be a good starting point for what we want to benchmark

vlandham:
this tool from Facebook should be useful for this timing process:

https://facebook.github.io/react/docs/perf.html


You should be able to attach it to `window` and then from the console:

```
window.Perf.start()
// load large file
window.Perf.stop()
window.Perf.printExclusive()
```

And get a lot of timing info.

PR #481, opened 2017-07-27
[WIP] Add benchmark middleware outside production

vlandham:
@kanitw - @tafsiri and i were pairing a bit on some ideas. 

Yannick detailed some things that could be benchmarked: https://github.com/vega/voyager/issues/385

this PR adds a tiny timer middleware on the redux store. 
wanted your thoughts on if this is moving in the right direction - or if we need to move this up towards UI rendering. 

additional directions we could take this middleware approach:

* filter so only certain events are recorded
* attempt to name match Result_request/receive - so we can auto aggregate these time values
* attempt to store the timing values themselves back into the datastore - so we could try to download them or view them in voyager.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -5,6 +5,7 @@ import {StateWithHistory} from 'redux-undo';
 
 import {DEFAULT_STATE, StateBase} from '../models';
 import {rootReducer} from '../reducers';
+import {benchmark} from './redux-benchmark';
 
 // Imports to satisfy --declarations build requirements
 // https://github.com/Microsoft/TypeScript/issues/9944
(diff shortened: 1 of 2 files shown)


vlandham:
interesting. 

the: 

> Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the Connect(BookmarkBase) component.

error showed up again here, and then went away....

vlandham:
screen shot 2017-07-27 at 10 38 53 am

vlandham:
closing as it is no longer relevant. See #385

PR #492, opened 2017-07-28
Split `selectors/index.ts` as it is becoming too big

kanitw:
cc: @starry97  

This will prevent the test file from being too bloating.


src/selectors/index.ts
```diff
@@ -1,122 +1,16 @@
...
-      return undefined;
-    }
-    return {
-      data: data,
-      transform: getTransforms(filters),
-      ...getTopSpecQueryItem(mainResult.modelGroup).spec
-    };
-  }
-);
+export * from './dataset';
```
kanitw: cc: @starry97 @felixcodes  -- if you are not familiar with the benefit of using `index.ts` -- basically we can split file and re-export.  

As a result, we don't really need to fix imports in other modules.

Issue #482, opened 2017-07-27
copy spec and bookmark icon size


starry97:
The color and size of the copy spec icon and bookmark icon is inconsistent.

![image](https://user-images.githubusercontent.com/19985016/28685865-0fc8488e-72be-11e7-8642-b9ff0493c0ff.png)

starry97:
closed via #497

PR #497, opened 2017-07-28
Add more padding to plot icons; change color to be consistent

starry97:
Related to #482. 

It does not change the size of icons, because the FontAwesome icons we use have different sizes. Adding more spaces make them look better.

### DIFF
--- a/src/components/plot/bookmarkbutton.scss
+++ b/src/components/plot/bookmarkbutton.scss
@@ -1,9 +1,16 @@
+@import './plot.scss';
+
 .bookmark-alert {
-    color: #fff;
-    background-color: #38425d;
-    padding: 8px;
-    font-size: 14px;
-    line-height: 18px;
-    min-width: 150px;
-    border-radius: 3px;
+  color: #fff;
+  background-color: #38425d;
+  padding: 8px;
+  font-size: 14px;
+  line-height: 18px;
+  min-width: 150px;
+  border-radius: 3px;
+}
+
+.bookmark-command-selected {
+  @extend .command;
+  color: #0c96d0;
 }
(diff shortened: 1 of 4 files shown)


src/components/plot/bookmarkbutton.tsx
```diff
@@ -42,7 +42,7 @@ export class BookmarkButtonBase extends React.PureComponent<BookmarkProps, Bookm
       >
         <i
           className="fa fa-bookmark command-bookmark"
-          style = {{color: bookmarkColor, paddingLeft: '3px'}}
+          style = {{color: bookmarkColor, paddingLeft: '12px', cursor: 'pointer'}}
```
kanitw: Instead of adding a custom `style`, we should add a class called `command`, which set `margin-left` (not padding), `cursor`, and `color` and use it for all command buttons in `<Plot>`

src/components/plot/bookmarkbutton.scss
```diff
@@ -1,9 +1,16 @@
+@import './plot.scss';
```
starry97: @kanitw Do you think it's better to import it or should I copy and paste the command class from plot.scss?

kanitw: I think import is better than copy and paste for sure :)

src/components/plot/plot.scss
```diff
@@ -37,12 +37,14 @@
   float: right;
 }
 
-.bookmark-button, .copy-button {
+.command {
   cursor: pointer;
-  margin-left: 4px;
+  margin-left: 20px;
```
kanitw: Why changing this to 20?   Can you share a screenshot?  In general, sharing a screenshot when you make style change is very helpful for review.  

cc: @felixcodes

starry97:
image

src/components/plot/plot.scss
starry97: I found myself accidentally click on the icon next to it, so I made the margin larger. 

Screenshot:
image

kanitw:
I think its too wide -- better use the old number from angular. But yeah 4
is too narrow

starry97:
screen shot 2017-07-31 at 10 45 30 am

PR #502, opened 2017-07-28
Use CompassQL v0.19.1 and update lock file

felixcodes:
@kanitw Do you want to bump and release Voyager as well?


kanitw:
No need.  Thanks for asking :)

Issue #478, opened 2017-07-26
React Modal CSS makes everything too big


felixcodes:
Everything in bookmark modal and dataset modal looks like its been magnified.

kanitw:
@felixcodes  -- You see -- we haven't closed this :p

PR #507, opened 2017-07-31
Fix size issue in Data Selector modal

felixcodes:
#478 

- Add `className='voyager'` to the Modal to resize everything back to normal.

### DIFF
--- a/src/components/data-pane/data-selector.tsx
+++ b/src/components/data-pane/data-selector.tsx
@@ -50,6 +50,7 @@ export class DataSelectorBase extends React.PureComponent<DataSelectorProps, any
          onRequestClose={this.closeModal}
          contentLabel="Data Selector"
          styleName="modal"
+         className="voyager"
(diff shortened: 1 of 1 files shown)


src/components/data-pane/data-selector.tsx
```diff
@@ -48,8 +48,9 @@ export class DataSelectorBase extends React.PureComponent<DataSelectorProps, any
         <Modal
          isOpen={this.state.modalIsOpen}
          onRequestClose={this.closeModal}
-         contentLabel="Data Selector"
-         styleName="modal"
+         contentLabel='Data Selector'
```
kanitw: Why changing to single quote?  We use double quote?

kanitw: @felixcodes Maybe it's worth researching appropriate TSLint flag (in another PR) to enforce consistency for this.

PR #509, opened 2017-07-31
Fix `copy-to-clipboard` plot command hitbox

felixcodes:
- Remove `<span>` surrounding the clipboard icon since it caused the clipboard's hitbox to extend onto the bookmark icon to its left. This would cause the spec to be copied when the right side of the bookmark icon was clicked (shown below).

Before:
image


kanitw:
Is this fixing #510, if so please close the issue accordingly.

PR #511, opened 2017-07-31
Decouple specPreview from Shelf

### LINKED ISSUE #505: Field added to the encoding preview cannot be removed
When hovering over the wildcard field in specified views, the field added to the encoding preview cannot be removed. At the same time, it triggers all plots to rerender.
bug

starry97:
Fix #505


src/actions/shelf-preview.ts
```diff
@@ -0,0 +1,12 @@
+import {FacetedCompositeUnitSpec} from 'vega-lite/build/src/spec';
+import {PlainReduxAction, ReduxAction} from './redux-action';
+
+export type ShelfPreviewAction = ShelfSpecPreview | ShelfSpecPreviewDisable;
+
+export const SHELF_SPEC_PREVIEW = 'SHELF_SPEC_PREVIEW';
```
kanitw: Maybe SHELF_PREVIEW_SPEC?  (So the state remains the prefix.)

starry97:
@kanitw Yes, this fixes #505

felixcodes:
@kanitw @starry97 
Just an FYI: we'll have to decouple this a step further later as the preview is added to the undo-redo stack.

kanitw:
@felixcodes Good catch.  Maybe add a TODO item to the decouple issue?

PR #515, opened 2017-08-01
Fix swapping shelf field bug

### LINKED ISSUE #503: Swapping shelf does not work
We also got this error:

```
Uncaught Error: Field dragged from unregistered source type to EncodingShelf
    at Object.drop (encoding-shelf.tsx:90)
    at Target.drop (createTargetFactory.js:80)
    at dragDrop.js:160
    at Array.forEach (<anonymous>)
    at DragDropManager.drop (dragDrop.js:157)
    at Object.drop (DragDropManager.js:92)
    at HTML5Backend.handleTopDrop (HTML5Backend.js:582)
```

starry97:
Fix #503

### DIFF
--- a/src/components/field/index.tsx
+++ b/src/components/field/index.tsx
@@ -49,7 +49,7 @@ export interface FieldPropsBase {
 export interface FieldProps extends FieldDragSourceProps, FieldPropsBase {};
 
 export interface FieldState {
-  popupIsOpened?: boolean;
+  popupIsOpened: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


kanitw:
Is this because the popup is preventing the drop?

starry97:
It shows an error cannot read 'popupIsOpened' of null when I tried to swap, which prevents the drop. It was caused because a function did not check whether the state exists, and I felt like having optional state is confusing, so I changed it.

kanitw:
Okie Thanks!

PR #516, opened 2017-08-02
Clicking on IDL logo opens another page

starry97:
This fixes the voyager-electron issue https://github.com/vega/voyager-electron/issues/19


kanitw:
Thanks :)

PR #519, opened 2017-08-02
Apply filters to plot list

### LINKED ISSUE #448: Filters are not applied to the plot gallery
- [x] We should consolidate a method that use to inject data and transforms into specs

starry97:
Fix #448


starry97:
@kanitw Is it necessary to write another function that inject filters to spec?

src/models/plot.ts
```diff
@@ -23,13 +24,26 @@ export interface PlotObject {
...
 
-export function extractPlotObjects(modelGroup: SpecQueryGroup<PlotObject>) {
+export function extractPlotObjects(modelGroup: SpecQueryGroup<PlotObject>,
+                                   filters: Array<RangeFilter|OneOfFilter>): PlotObject[] {
   return modelGroup.items.map(item => {
     if (isSpecQueryGroup<PlotObject>(item)) {
       const childModelGroup = item as SpecQueryGroup<PlotObject>;
-      return getTopSpecQueryItem(childModelGroup);
+      return {
+        ...getTopSpecQueryItem(childModelGroup),
```
kanitw: this seems redundant to call twice?

Let's make a method that takes a spec and filters as input can combine them and re-use it in both cases.

PR #526, opened 2017-08-03
Fix moving encoding field

starry97:
Previously there was a bug preventing the encoding field being moved to another shelf, now this is fixed.

### DIFF
--- a/src/components/encoding-pane/encoding-shelf.tsx
+++ b/src/components/encoding-pane/encoding-shelf.tsx
@@ -30,7 +30,7 @@ export interface EncodingShelfPropsBase extends ActionHandler<ShelfEncodingActio
 
   fieldDef: ShelfFieldDef;
 
-  schema?: Schema;
+  schema: Schema;
 }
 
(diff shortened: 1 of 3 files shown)


src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -28,6 +29,8 @@ export interface EncodingShelfPropsBase extends ActionHandler<ShelfEncodingActio
   id: ShelfId;
 
   fieldDef: ShelfFieldDef;
+
+  schema?: Schema;
```
kanitw: Why is this optional?

starry97: Because if it's wildcard fields or Number of Record field, they don't need the schema since they cannot have filters.

starry97: Oh but I should have set it to a required parameter, just don't call schema.domain to get the filters.

starry97: Maybe we should make it required since Field component in encoding shelf does not differentiate wildcard fields from other fields

PR #529, opened 2017-08-04
Indicate users cannot drop wildcard or number of records to filter pane

starry97:


kanitw:
screenshot? :)

starry97:
drop

kanitw:
Can we say "Cannot drop a wildcard field here" and "Cannot drop a count field here" so it's clearer?

PR #535, opened 2017-08-04
Use empty dataset as default app state

### LINKED ISSUE #442: Better default state of the program
Right now we load a weird mock dataset by default.

starry97:
Fix #442
image


kanitw:
- Looks like test is now failing  
- Can we also hide Fields and Wildcard Fields header?

PR #538, opened 2017-08-07
Use CQL 0.19.4 and update lock file and tests

felixcodes:


felixcodes:
Passes locally. We'll merge and ignore all PRs that pass locally but fail on Travis with `A worker process has quit unexpectedly! Most likely this an initialization error.` for now.

PR #541, opened 2017-08-07
Upgrade CQL & Fix Long standing bugs

kanitw:


kanitw:
Please only approve / comment -- don't rebase merge this branch as I have already merge it to another branch -- so we have to manually merge this.

src/components/app.test.ui.tsx
```diff
@@ -133,7 +133,7 @@ describe('Voyager', () => {
               const fieldList = wrapper.find('.encoding-shelf__encoding-shelf');
               const fields = fieldList.map(d => d.text());
 
-              expect(fields).toContain('x close');
+              expect(fields).toContain('x binclose');
```
starry97: Shouldn't the spec be changed to have field 'x binclose' instead of 'x close'?

kanitw: the spec is binned, and this text is from the UI with bin(close)  but parenthesis are displayed using before: and after: element so they are not included here.

PR #544, opened 2017-08-07
Fix bookmark notes css

### LINKED ISSUE #518: The height of notes input is small when the plot vertically overflows
screen shot 2017-08-02 at 2 32 09 pm
screen shot 2017-08-02 at 2 35 20 pm

starry97:
Make bookmark notes style consistent with the old app.
Fix #518 
image


kanitw:
Thanks!

PR #546, opened 2017-08-08
Set default state to be empty dataset

kanitw:


src/components/app.test.ui.tsx
```diff
@@ -87,24 +88,30 @@ describe('Voyager', () => {
...
+      const data: Data = {values};
       const spec: Object = {
         "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
         "data": {
           "values": [
-            {"date": "24-Apr-07", "close": "93.24"},
-            {"date": "25-Apr-07", "close": "95.35"},
-            {"date": "26-Apr-07", "close": "98.84"},
-            {"date": "27-Apr-07", "close": "99.92"}
+            {"date": "24-Apr-07", "binclose": "93.24"},
```
kanitw: Why changing this?   -- close is already right and the field is binned in the spec

src/lib-voyager.test.ui.tsx
```diff
@@ -220,17 +220,17 @@ describe('lib-voyager', () => {
             "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
             "data": {
               "values": [
-                {"date": "24-Apr-07", "close": "93.24"},
-                {"date": "25-Apr-07", "close": "95.35"},
-                {"date": "26-Apr-07", "close": "98.84"},
-                {"date": "27-Apr-07", "close": "99.92"}
+                {"date": "24-Apr-07", "binclose": "93.24"},
```
kanitw: same here

src/lib-voyager.tsx
```diff
@@ -77,7 +77,8 @@ export class Voyager {
    * @memberof Voyager
    */
   public setSpec(spec: Object) {
-    this.data = undefined;
+    const data = "data";
+    this.data = spec[data];
```
kanitw: Why don't just this.data = `spec.data` or `spec['data']` ?

kanitw: I think we should track where they do JSON validation and consider if we should move the logic to here.

src/reducers/dataset.test.ts
```diff
@@ -3,20 +3,20 @@ import {FieldSchema, Schema} from 'compassql/build/src/schema';
...
         DATASET_URL_RECEIVE, DATASET_URL_REQUEST} from '../actions/dataset';
-import {Dataset, DEFAULT_DATASET} from '../models/dataset';
+import {Dataset, SAMPLE_DATASET} from '../models/dataset';
 import {datasetReducer} from './dataset';
 
 describe('reducers/dataset', () => {
   describe(DATASET_URL_REQUEST, () => {
     it('returns new dataset state with isLoading = true', () => {
-      expect(datasetReducer(DEFAULT_DATASET, {
+      expect(datasetReducer(SAMPLE_DATASET, {
```
kanitw: Why do we have to keep the sample dataset here?

src/lib-voyager.test.ui.tsx
```diff
@@ -93,11 +93,7 @@ describe('lib-voyager', () => {
           setTimeout(() => {
             try {
               const fieldList = document.querySelectorAll('.field-list__field-list-item');
-              const fields = Array.prototype.map.call(fieldList, (d: Node) => d.textContent);
-
-              expect(fields).toContain(' q1');
-              expect(fields).toContain(' q2');
-
+              expect(fieldList.length).toEqual(0);
```
kanitw: It seems like the description of this test doesn't really what the test does. 
(It tests two things -- maybe it "initializes with empty data and can be updated with cusomized data"?)

src/components/data-pane/data-selector.tsx
```diff
@@ -190,7 +190,7 @@ export const DataSelector = connect(
   (state: State) => {
     return {
       data: state.present.dataset,
-      name: state.present.dataset.name,
+      name: state.present.dataset.name
```
kanitw: Please undo the change to this line and squash it in to this (3f0fd2) -- otherwise you'll cause a merge conflict.

src/reducers/dataset.test.ts
kanitw: If possible, maybe we can even get rid of it if it is only used in this file.

src/components/data-pane/index.tsx
```diff
@@ -20,26 +20,31 @@ export interface DataPanelProps extends ActionHandler<DatasetAsyncAction | Shelf
 export class DataPanelBase extends React.PureComponent<DataPanelProps, {}> {
   public render() {
     const {name} = this.props.data;
+    const fieldLength = this.props.data.schema.fieldSchemas.length;
```
kanitw: fieldCount?

src/lib-voyager.tsx
```diff
@@ -77,8 +81,24 @@ export class Voyager {
    * @memberof Voyager
    */
   public setSpec(spec: Object) {
-    const data = "data";
-    this.data = spec[data];
+
+    const ajv = new Ajv({
```
kanitw: Hmm -- don't we already do this somewhere? 

I mean it's good that we do here but I wanna make sure that we don't run it twice.

src/components/app.tsx
```diff
@@ -1,41 +1,30 @@
...
-import {EncodingPane} from './encoding-pane';
-import {Header} from './header';
-import {LoadData} from './load-data-pane/index';
-import {ViewPane} from './view-pane';
+import {StateBase} from '../models/index';
+import {AppRoot} from './app-root';
+import './app.scss';
 
 
 export type VoyagerData = Data;
```
kanitw: Is this line used at all?

src/components/app.tsx
```diff
@@ -167,4 +137,4 @@ class AppBase extends React.PureComponent<Props, {}> {
   }
 }
 
-export const App = DragDropContext(HTML5Backend)(AppBase);
+export const App = AppBase;
```
kanitw: There is no wrapping anymore.  Get rid of this line and just rename AppBase to App?

src/components/app.tsx
starry97: Yes, type VoyagerData is used in lib-voyager.tsx

kanitw: Wonder if we better remove it since this is not consistently used anyway?

src/lib-voyager.test.ui.tsx
starry97: Done

src/reducers/dataset.test.ts
```diff
@@ -1,22 +1,74 @@
...
+import {FieldSchema, PrimitiveType, Schema} from 'compassql/build/src/schema';
 
 import {DATASET_INLINE_RECEIVE, DATASET_SCHEMA_CHANGE_FIELD_TYPE, DATASET_SCHEMA_CHANGE_ORDINAL_DOMAIN,
         DATASET_URL_RECEIVE, DATASET_URL_REQUEST} from '../actions/dataset';
-import {Dataset, SAMPLE_DATASET} from '../models/dataset';
+import {Dataset} from '../models/dataset';
 import {datasetReducer} from './dataset';
 
+
+const sampleDataset: Dataset = {
```
kanitw: Do we really need sample dataset here?  Can't we use default dataset?

Issue #483, opened 2017-07-27
Decouple bookmark + shelf preview from Undoable State


felixcodes:
We need to separate persistent state and undoable state. Bookmark should exist in persistent state. 

+ shelf preview as well

PR #547, opened 2017-08-08
Decouple persistent state and undoable state

felixcodes:
#483 

- Separate persistent state from undoable state


src/components/data-pane/data-selector.tsx
```diff
@@ -188,9 +188,10 @@ const DataSelectorRenderer = CSSModules(DataSelectorBase, styles);
 
 export const DataSelector = connect(
   (state: State) => {
+    const presentUndoableState = state.undoable.present;
     return {
-      data: state.present.dataset,
-      name: state.present.dataset.name,
+      data: presentUndoableState.dataset,
```
kanitw: We should use a selector for dataset for consistency and for reduce the amount of updates needed if we reshape the state in the future.

src/components/data-pane/data-selector.tsx
```diff
@@ -188,9 +188,10 @@ const DataSelectorRenderer = CSSModules(DataSelectorBase, styles);
...
+      name: presentUndoableState.dataset.name,
```
kanitw: It's weird that we need name since it's already a part of data.  File an issue and submit a separate PR?

src/components/data-pane/index.tsx
```diff
@@ -47,9 +47,10 @@ export class DataPanelBase extends React.PureComponent<DataPanelProps, {}> {
 
 export const DataPane = connect(
   (state: State) => {
+    const presentUndoableState = state.undoable.present;
     return {
-      data: state.present.dataset,
-      config: state.present.config
+      data: presentUndoableState.dataset,
```
kanitw: We should use a selector for dataset, config for consistency and for reduce the amount of updates needed if we reshape the state in the future.

src/components/encoding-pane/index.tsx
```diff
@@ -162,12 +162,13 @@ class EncodingPanelBase extends React.PureComponent<EncodingPanelProps, {}> {
 
 export const EncodingPane = connect(
   (state: State) => {
+    const presentUndoableState = state.undoable.present;
     return {
-      spec: state.present.shelf.spec,
-      filters: state.present.shelf.spec.filters,
-      schema: state.present.dataset.schema,
+      spec: presentUndoableState.shelf.spec,
```
kanitw: We should use a selector for dataset, shelf, shelfPreview for consistency and for reduce the amount of updates needed if we reshape the state in the future.

src/models/index.ts
```diff
@@ -63,31 +75,52 @@ export interface SerializableState {
...
+  const persistentState = state.persistent;
+  const undoableState = state.undoable.present;
+
   const asSerializable = {
-    bookmark: state.bookmark,
-    config: state.config,
-    shelf: state.shelf,
-    shelfPreview: state.shelfPreview,
-    result: state.result,
+    bookmark: persistentState.bookmark,
```
kanitw: just return it, no need for const

src/models/index.ts
```diff
@@ -63,31 +75,52 @@ export interface SerializableState {
...
-export function fromSerializable(serializable: SerializableState): Readonly<StateBase> {
-  // We make a clone of this object to not modify the input param.
-  const deserialized = duplicate(serializable) as any;
+export function fromSerializable(serializable: SerializableState): Readonly<State> {
+  const {bookmark, config, shelf, shelfPreview, result, dataset, tableschema} = serializable;
 
-  // Add a schema object with a hydrated version of the table schema
-  deserialized.dataset.schema = new Schema(deserialized.tableschema);
-  delete deserialized.dataset.tableschema;
+  const deserialized: State = {
```
kanitw: just return it, no need for const

src/selectors/shelf.test.ts
```diff
@@ -8,45 +8,36 @@ describe('selectors/shelf', () => {
...
-            spec: {
-              ...DEFAULT_STATE.shelf.spec,
-              filters
-            }
-          }
-        }
-      };
+
+      const state = {... DEFAULT_STATE};
+      state.undoable.present.shelf.spec.filters = filters;
```
kanitw: This is causing side effect to DEFAULT_STATE and may make other tests fail!

src/components/data-pane/data-selector.tsx
felixcodes: update: we fixed this in one of the commits in this PR (to be pushed)

src/components/header/bookmark.tsx
```diff
@@ -134,7 +134,7 @@ const BookmarkRenderer = CSSModules(BookmarkBase, styles);
 export const BookmarkPane = connect(
   (state: State) => {
     return {
-      bookmark: state.present.bookmark
+      bookmark: state.persistent.bookmark
```
kanitw: selectBookmark?

src/selectors/result.ts
```diff
@@ -1,31 +1,30 @@
...
 export const selectResult: {
-  [k in ResultType]?: Selector<StateWithHistory<StateBase>, Result>
+  [k in ResultType]?: Selector<State, Result>
 } = RESULT_TYPES.reduce((selectors, resultType) => {
-  selectors[resultType] = (state: State) => state.present.result[resultType];
+  selectors[resultType] = (state: State) => state.undoable.present.result[resultType];
   return selectors;
 }, {});
 
+// const selectMainResult = (state: State) => state.undoable.present.result.main;
```
kanitw: Why adding this unused line?

PR #551, opened 2017-08-08
Remove Number of Record from count field

### LINKED ISSUE #543: Count Field now shows both COUNT and Number of Records
Let's just show one of them.

kanitw:
Fix #543


kanitw:
Nice now!

voyager_2

PR #553, opened 2017-08-09
Logic for supporting wildcard function in the shelf

kanitw:

### DIFF
--- a/src/models/shelf/function.ts
+++ b/src/models/shelf/function.ts
@@ -34,6 +34,20 @@ const TEMPORAL_FUNCTIONS = [
   'seconds', 'milliseconds'
 ];
 
+const FUNCTIONS_INDEX: {
+  [K in ShelfFunction]? : number;
+} = {
+  ...QUANTITATIVE_FUNCTIONS.reduce((index, fn, i) => {
+    index[fn] = i;
+    return index;
+  }, {}),
+
+  ...TEMPORAL_FUNCTIONS.reduce((index, fn, i) => {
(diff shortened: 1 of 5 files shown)


src/models/shelf/function.ts
```diff
@@ -0,0 +1,158 @@
...
+      fns.unshift(undefined);
+    }
+    return {enum: sortFunctions(fns)};
+  }
+}
+
+export function sortFunctions(fns: ShelfFunction[]): ShelfFunction[] {
+  // Sort functions to match order in the the UI
+  // FIXME: implement the sort
+  return fns;
```
kanitw: Still need to sort based on the function order.

src/models/shelf/function.ts
```diff
@@ -0,0 +1,158 @@
...
+    return false;
+  }
+  return fn !== undefined && fn !== null;
+}
+
+export function fromFieldQueryFunctionMixins(
+  fieldQParts: Pick<FieldQuery, 'aggregate' | 'timeUnit' | 'bin' | 'hasFn'>
+): ShelfFunction | Wildcard<ShelfFunction> {
+
+  // FIXME make this a parameter
```
kanitw: Please file an issue that we should make this a parameter.  (Let's do it later.)

kanitw:
The logic here should be **mostly** correct -- but I haven't written any test though.  
So write some and feel free to discuss why I do certain things

kanitw:
@felixcodes  This should be ready to go!

PR #567, opened 2017-08-10
[WIP] Remove app-root

### LINKED ISSUE #561: Get rid of App Root
We can use `connect` with app, we just need to set its generic interface correctly.

starry97:
Fix #561


src/components/app.tsx
```diff
@@ -112,3 +148,11 @@ export class App extends React.PureComponent<Props, {}> {
     });
   }
 }
+
+export const App = connect<ConnectProps, null, OwnProps>(
+  (state: State) => {
+    return {
+      connectData: selectData(state)
```
kanitw: data and connectData are conflicting -- maybe it's still not a bad idea to split app-root?

kanitw:
Let s keep it open so we can reconsider  it one more time (with less time pressure)

kanitw:
Let's keep it as is

PR #570, opened 2017-08-10
Display dataset filename when loading from CSV/TSV/JSON data or Vega-Lite spec in Voyager-Electron

felixcodes:
Related Voyager-Electron issue: https://github.com/vega/voyager-electron/issues/18
Related Voyager-Electron PR: https://github.com/vega/voyager-electron/pull/28

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -10,9 +10,14 @@ import {VoyagerConfig} from '../models/config';
 import {State} from '../models/index';
 import {AppRoot} from './app-root';
 
+export interface VoyagerData {
+  data: Data;
+  filename: string;
+}
+
 export interface Props extends React.Props<App> {
   config?: VoyagerConfig;
-  data?: Data;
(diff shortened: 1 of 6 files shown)


felixcodes:
Data pane now looks like this when loading from Vega-Lite spec:

image

and like this when loading from CSV (and TSV and JSON):

image

src/components/app.tsx
```diff
@@ -10,11 +10,21 @@ import {VoyagerConfig} from '../models/config';
...
+
+export interface SpecWithFilename {
+  spec: TopLevel<FacetedCompositeUnitSpec>;
+  filename: string;
+}
+
 export interface Props extends React.Props<App> {
   config?: VoyagerConfig;
-  data?: Data;
+  data?: VoyagerData;
```
kanitw: Instead of making two properties having "filename", why don't we make filename a new prop of the application?

PR #597, opened 2017-08-10
Display errors and warnings 

### LINKED ISSUE #588: Error handling mechanism
- Load 
  - [ ] Errors
  - [ ] warnings (we can drop something)

- Spec / Plot / Vega-Lite 
   - [ ] For each plot, add logs / warning that are unique to the store 
   - [ ] Whenever the spec changes, clear the log for shelf.

- Filter 
  - [ ] Warn that they cannot drop two filters to the same field. (Show alert)

- ~~Bugs~~ 
  - [x] ~~Undefined Error, etc.  (These are usually disastrous) -- can we design in a way that users can continue.~~

Two options:
1) Distribute error to different parts of the state
    -  Easy to make all other shelf actions clear the error

2) Centralize and have separate keys
    - Easy to combine them in the UI

starry97:
Fix #588

- [x] log UI renders log in the redux store
- [x] Add log in the redux store
- [x] Add localLogger and wire it to plot

### DIFF
--- a/src/actions/dataset.ts
+++ b/src/actions/dataset.ts
@@ -9,6 +9,7 @@ import {fetchCompassQLBuildSchema} from '../api/api';
 import {State} from '../models/index';
 import {selectConfig} from '../selectors';
 import {Action} from './index';
+import {LOG_ERRORS_ADD} from './log';
 import {ReduxAction} from './redux-action';
 import {RESET} from './reset';
 
@@ -79,7 +80,12 @@ export function datasetLoad(name: string, dataset: Data): DatasetLoad {
           dispatch(ActionCreators.clearHistory());
         });
     } else {
-      throw new Error('dataset load error: dataset type not detected');
+      dispatch({
+        type: LOG_ERRORS_ADD,
+        payload: {
+          errors: ['dataset load error: dataset type not detected']
+        }
+      });
     }
 
 
--- a/src/actions/dataset.ts
+++ b/src/actions/dataset.ts
@@ -9,7 +9,7 @@ import {fetchCompassQLBuildSchema} from '../api/api';
 import {State} from '../models/index';
 import {selectConfig} from '../selectors';
 import {Action} from './index';
-import {LOG_ERRORS_ADD} from './log';
+import {LOG_ERROR_CHANGE} from './log';
 import {ReduxAction} from './redux-action';
 import {RESET} from './reset';
 
@@ -58,8 +58,24 @@ export function datasetLoad(name: string, dataset: Data): DatasetLoad {
       const url = dataset.url;
 
       return fetch(url)
-        .then(response => response.json()) // TODO: handle error
-        .then(data => fetchCompassQLBuildSchema(data, config)) // TODO: handle error
+        .then(response => response.json())
+        .catch(err => {
+          dispatch({
+            type: LOG_ERROR_CHANGE,
+            payload: {
+              error: err
+            }
+          });
+        })
+        .then(data => fetchCompassQLBuildSchema(data, config))
+        .catch(err => {
+          dispatch({
+            type: LOG_ERROR_CHANGE,
+            payload: {
+              error: err
+            }
+          });
+        })
         .then(schema => {
 
           dispatch({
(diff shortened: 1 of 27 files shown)


starry97:
screen shot 2017-08-21 at 3 05 02 pm
screen shot 2017-08-21 at 2 46 46 pm

src/reducers/shelf/filter.ts
```diff
@@ -19,7 +19,8 @@ export function filterReducer(shelfSpec: Readonly<ShelfUnitSpec> = DEFAULT_SHELF
       const {filter} = action.payload;
       let index = action.payload.index;
       if (contains(shelfSpec.filters, filter)) {
-        throw new Error('Cannot add more than one filter of the same field');
+        window.alert('Cannot add more than one filter of the same field');
```
kanitw: Hmm, the reducer shouldn't cause side effect to the UI (via `window`). 

Instead, I think throwing error here is totally fine.  

What we need to design is where do we catch these errors and correctly throw error.

kanitw:
Thanks for the PR.  I think this is a good start.  

From looking at code quickly, it seems like there are multiple ways that we handle error.  

After handling issue, could you please summarize what are the mechanism that we have? 
With that I'll be able to give a better higher level comment :)

src/actions/dataset.ts
```diff
@@ -79,7 +80,12 @@ export function datasetLoad(name: string, dataset: Data): DatasetLoad {
           dispatch(ActionCreators.clearHistory());
         });
     } else {
-      throw new Error('dataset load error: dataset type not detected');
+      dispatch({
+        type: LOG_ERRORS_ADD,
+        payload: {
+          errors: ['dataset load error: dataset type not detected']
```
kanitw: Dataset load

src/reducers/shelf/filter.ts
kanitw: Reading more, throwing error in reducer is not a good idea either.

Instead, we may want to move the validation logic to the relevant filter components

src/store/index.ts
```diff
@@ -28,8 +28,9 @@ if (process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test') {
   middleware.push(loggerMiddleware);
 }
 
+export const localLogger = new Logger();
```
kanitw: I don't think we should have a singleton local logger.  This makes the VegaLite unnecessary coupled.

src/components/vega-lite/index.tsx
```diff
@@ -3,16 +3,19 @@ import * as vega from 'vega';
 import * as vl from 'vega-lite';
 import {TopLevelExtendedSpec} from 'vega-lite/build/src/spec';
 import * as vegaTooltip from 'vega-tooltip';
+import {LOG_ERROR_CHANGE, LOG_WARNINGS_ADD, LogAction} from '../../actions/log';
+import {ActionHandler} from '../../actions/redux-action';
+import {localLogger} from '../../store/index';
```
kanitw: I don't think we should have a singleton local logger.  This makes the VegaLite unnecessary coupled.  Instead, each VegaLite component takes its own logger as input.

src/components/vega-lite/index.tsx
```diff
@@ -3,16 +3,19 @@ import * as vega from 'vega';
...
 
-export interface VegaLiteProps {
+export interface VegaLiteProps extends ActionHandler<LogAction> {
```
kanitw: Update: I would prefer having ~~`warn` and `error` methods as~~
a logger as `VegaLiteProps` and let each logger in `Plot` calls handleAction instead.  (You can name it `plotLogger`)

src/models/logger.ts
```diff
@@ -0,0 +1,37 @@
+export class Logger {
```
kanitw: Why do we need this class at all though?  I think we can implement a plotLogger as a part of log, which in turn trigger a relevant LogAction directly.

src/reducers/log.test.ts
```diff
@@ -0,0 +1,62 @@
+import {LOG_ERROR_CHANGE, LOG_ERROR_REMOVE, LOG_WARNINGS_ADD, LOG_WARNINGS_REMOVE} from '../actions/log';
+import {DEFAULT_LOG, Log} from '../models/log';
+import {logReducer} from './log';
+
+const simpleLog: Log = {
```
kanitw: Maybe move this down to where it's used?

src/components/vega-lite/index.tsx
```diff
@@ -23,8 +26,37 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
...
+    //       {"a": "I", "b": 52}
+    //     ]
+    //   },
+    //   "mark": "bar",
+    //   "encoding": {
+    //     "x": {"field": "a", "type": "quantitative"},
+    //     "y": {"field": "b", "type": "quantitative"}
+    //   }
+    // };
+    const {spec} = vl.compile(vlSpec, localLogger);
```
kanitw: Wrap this with try catch

starry97:
Open #650 instead

PR #611, opened 2017-08-12
Make filter pane scrollable

### LINKED ISSUE #591: Make Encoding/Filter Panel Scrollable
When adding multiple filter panels, they can go offscreen and there is no way to sort the encoding/filter panel.

starry97:
Fix #591


starry97:
image

PR #614, opened 2017-08-15
Fix function picker so `-` displays as checked on initial load and click

felixcodes:
Fix function selector radios so that it shows `-` as selected when it's clicked

image


src/components/encoding-pane/function-picker.tsx
```diff
@@ -28,18 +28,24 @@ export class FunctionPickerBase extends React.PureComponent<FunctionPickerProps,
...
-          value={f || '-'}
-          checked={f === fn}
-          onChange={this.onFunctionChange}
-        />
-        {' '}
-        {f || '-'}
-      </label>
-    ));
+    const radios = supportedFns.map(f => {
+      if (f === undefined && fn !== undefined) {
```
kanitw: Make it work with out `fn !== undefined`

PR #640, opened 2017-08-18
Add `index` and `limit` properties required by StateWithHistory

felixcodes:


src/store/index.ts
```diff
@@ -39,6 +39,8 @@ export function configureStore(initialState = DEFAULT_STATE) {
       future: [],
       _latestUnfiltered: null,
       group: null,
+      index: null,
+      limit: 30
```
kanitw: Isn't it weird that we have to modify this twice?  Why do we need `initialStateWithHistory` at all? 

Isn't `initialState` sufficient?

PR #650, opened 2017-08-22
Warning and error log reducer & UI

starry97:
Part of #588


src/components/plot/index.tsx
```diff
@@ -125,17 +131,35 @@ export class PlotBase extends React.PureComponent<PlotProps, PlotState> {
...
         >
-          <VegaLite spec={spec}/>
+          <VegaLite spec={spec} logger={this.plotLogger}/>
         </div>
         {notesDiv}
       </div>
     );
   }
 
+  protected componentDidMount() {
```
kanitw: Why do we need to flush this during "componentDidMount()"?  

Plus, I thought we agree that we can just create one log action per warning.  

That said, if there is a good solution, I am okay with flushing them all at once, but I'm not sure if this is the right trigger.  Isn't componentDidMount triggered only when it is mounted the first time?

Since it doesn't seem right this gonna be correct, why don't we just make logger trigger `handleAction`?

src/actions/index.ts
```diff
@@ -58,6 +61,11 @@ const ACTION_TYPE_INDEX: {[k in ActionType]: 1} = {
   FILTER_MODIFY_ONE_OF: 1,
   FILTER_REMOVE: 1,
 
+  LOG_ERRORS_ADD: 1,
+  LOG_ERRORS_REMOVE: 1,
```
kanitw: When is error moved and warnings not removed? 

I wonder if we should just have a single `LOG_CLEAR` action?

src/reducers/log.ts
```diff
@@ -0,0 +1,42 @@
...
+      warnings: []
+    };
+  }
+
+  switch (action.type) {
+    case LOG_ERRORS_ADD:
+      const {errors} = action.payload;
+      return {
+        ...log,
+        errors: addNoDuplicateArrays(log.errors, errors)
```
kanitw: This will always re-create log object no matter we have new error added or not.

src/reducers/log.ts
```diff
@@ -0,0 +1,42 @@
...
+      };
+    case LOG_WARNINGS_ADD:
+      const {warnings} = action.payload;
+      return {
+        ...log,
+        warnings: addNoDuplicateArrays(log.warnings, warnings)
```
kanitw: This will always re-create log object no matter we have new warning added or not.

src/reducers/log.ts
kanitw: Instead, I would do something like 
```
const newUniqueErrors = errors.filter(e => !contains(log.errors, e));
if (newUniqueErrors) {
  return  {
    ...log,
    errors: log.error.concat(newUniqueErrors)
  };
}
return log; 
```

src/selectors/shelf.test.ts
```diff
@@ -21,6 +22,9 @@ describe('selectors/shelf', () => {
           config: {
             ...DEFAULT_VOYAGER_CONFIG
           },
+          log: {
```
kanitw: Why don't we just make `persistent: DEFAULT_PERSISTENT_STATE`?

Seems like we're not adding anything here anyway.

src/actions/dataset.ts
```diff
@@ -57,8 +57,14 @@ export function datasetLoad(name: string, dataset: Data): DatasetLoad {
       const url = dataset.url;
 
       return fetch(url)
-        .then(response => response.json()) // TODO: handle error
-        .then(data => fetchCompassQLBuildSchema(data, config)) // TODO: handle error
+        .then(response => response.json())
+        .catch(err => {
```
kanitw: `.catch(errorCatch)` is sufficient

src/reducers/shelf/filter.ts
```diff
@@ -89,14 +76,9 @@ export function filterReducer(shelfSpec: Readonly<ShelfUnitSpec> = DEFAULT_SHELF
...
       const modifyMinBound = (filter: RangeFilter) => {
-        const range = filter.range;
-        const maxBound = range[range.length - 1];
-        if (minBound > maxBound) {
-          throw new Error ('Minimum bound cannot be greater than maximum bound');
-        }
         return {
           ...filter,
-          range: [minBound, maxBound]
+          range: [minBound, filter.range[filter.range.length - 1]]
```
kanitw: Isn't the length always 2?  Can't we do `filter.range[1]`?

src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -134,6 +137,7 @@ class EncodingShelfBase extends React.PureComponent<EncodingShelfProps, {}> {
       <div styleName='field-wrapper'>
         <Field
           draggable={true}
+          filters={filters}
```
kanitw: Why does field inside encoding-shelf needs an array of filter?  This seems like an overkill.

src/components/field/index.tsx
```diff
@@ -294,17 +296,18 @@ export interface DraggedFieldIdentifier {
   fieldDef: ShelfFieldDef;
   parentId: FieldParentId;
   filter: RangeFilter | OneOfFilter;
+  filters: Array<RangeFilter | OneOfFilter>;
```
kanitw: Why do we need the whole filters array when we drag only one thing? This seems pretty coupled?

src/actions/log.ts
```diff
@@ -0,0 +1,19 @@
...
+export type LogErrorsAdd = ReduxAction<typeof LOG_ERRORS_ADD, {
+  errors: string[]
+}>;
+
+export const LOG_WARNINGS_ADD = 'LOG_WARNINGS_ADD';
+export type LogWarningsAdd = ReduxAction<typeof LOG_WARNINGS_ADD, {
+  warnings: string[]
+}>;
+
+export const LOG_WARNINGS_REMOVE = 'LOG_WARNINGS_REMOVE';
```
kanitw: `LOG_WARNINGS_ADD` has parameter, but LOG_WARNINGS_REMOVE does not.  Maybe LOG_WARNINGS_CLEAR is a better name?    Same for ERROR.   

I also wonder if we really need two separate CLEAR though?  If they really wanna dismiss warnings and errors, shouldn't we dismiss them in one click.

src/components/field/index.tsx
```diff
@@ -294,17 +296,18 @@ export interface DraggedFieldIdentifier {
...
   beginDrag(props): DraggedFieldIdentifier {
-    const {fieldDef, parentId, schema} = props;
+    const {fieldDef, parentId, schema, filters} = props;
     let domain;
     if (!isWildcard(fieldDef.field) && fieldDef.field !== '*') {
       domain = schema.domain({field: fieldDef.field});
     }
     const filter = getFilter(fieldDef, domain);
-    return {fieldDef, parentId, filter};
+    return {fieldDef, parentId, filter, filters};
```
kanitw: Why do we need to add `filters` here?

src/components/encoding-pane/filter-pane.tsx
```diff
@@ -205,9 +205,14 @@ const filterShelfTarget: DropTargetSpec<FilterPaneProps> = {
     if (monitor.didDrop()) {
       return;
     }
-    const {filter} = monitor.getItem() as DraggedFieldIdentifier;
+    const {filter, filters} = monitor.getItem() as DraggedFieldIdentifier;
```
kanitw: `filters` is already a prop of FilterPane.  Why do you need to get it from the dragged field, which semantically should be a single filter that you add.

src/components/field/index.tsx
```diff
@@ -27,6 +28,8 @@ export interface FieldDragSourceProps {
 }
 
 export interface FieldPropsBase {
+  filters?: Array<RangeFilter | OneOfFilter>;
```
kanitw: `filters` are only need if `filterShow` is true.  So you don't really need `filters` in many instances of Field. 

If you want to keep limit one filter per one field, One design option that would make the parameters of Field less exploded.  
```
filterShow?: boolean
```
to be

```
filterShow?: {
   filters:  Array<RangeFilter | OneOfFilter>
} 
```

This way it will be pretty clear that if you want to filter, pass in `filters` as it's needed for filter.

src/actions/log.ts
starry97: Right now warnings and errors are displayed in two separate pane (one in red and one in yellow). Do you think we should show them in one pane instead?

src/actions/dataset.ts
```diff
@@ -81,7 +90,9 @@ export function datasetLoad(name: string, dataset: Data): DatasetLoad {
     } else {
       throw new Error('dataset load error: dataset type not detected');
     }
-
-
   };
 };
+
+function errorCatch(err: string) {
+  window.alert(err);
```
kanitw: Is error always already a string?

starry97:
Updated UI image

src/models/logger.ts
```diff
@@ -0,0 +1,56 @@
+
+import {LOG_ERRORS_ADD, LOG_WARNINGS_ADD, LogAction} from '../actions/log';
```
kanitw: Maybe logger shouldn't belong in model?

PR #658, opened 2017-08-23
Data binding

### LINKED ISSUE #652: Use Data Binding
Instead of setting data to all specs, let's bind data at runtime

- [x] Make spec do not contain inline data / url data (use named data source instead)

- Bind data at runtime
  - [x] Add `data` as a prop to VegaLite module
  - [x] Correctly call data binding accordingly

kanitw:
This is the start for binding data at runtime (for @starry97  to take over)

Fix #652

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -6,7 +6,7 @@ import {Query} from 'compassql/build/src/query/query';
 import {recommend} from 'compassql/build/src/recommend';
 import {build as buildSchema, Schema} from 'compassql/build/src/schema';
 import 'isomorphic-fetch';
-import {Data} from 'vega-lite/build/src/data';
+import {Data, isUrlData} from 'vega-lite/build/src/data';
 import {VoyagerConfig} from '../models/config';
 import {fromSpecQueryModelGroup, ResultPlotWithKey} from '../models/result';
 
@@ -42,6 +42,11 @@ export function fetchCompassQLRecommend(query: Query, schema: Schema, data: Data
     return new Promise(resolve => {
       const modelGroup = recommend(query, schema).result;
 
+      // TODO:
+      // - replace this with different cached data source's unique names
+      // once we have multiple cached data source from Leilani's optimizer engine
+      // - (short term) always bind data at runtime
+      data = isUrlData(data) ? data : {name: 'source'};
       resolve(fromSpecQueryModelGroup(modelGroup, data));
     });
   }
--- a/src/components/plot/index.tsx
+++ b/src/components/plot/index.tsx
@@ -1,7 +1,9 @@
 import * as React from 'react';
 import * as CopyToClipboard from 'react-copy-to-clipboard';
 import * as CSSModules from 'react-css-modules';
+import {connect} from 'react-redux';
 import * as TetherComponent from 'react-tether';
+import {Data} from 'vega-lite/build/src/data';
 import {isDiscrete, isFieldDef} from 'vega-lite/build/src/fielddef';
 import {SortField, SortOrder} from 'vega-lite/build/src/sort';
 import {FacetedCompositeUnitSpec} from 'vega-lite/build/src/spec';
@@ -13,14 +15,16 @@ import {ShelfAction, SPEC_LOAD} from '../../actions/shelf';
 import {SHELF_PREVIEW_DISABLE, SHELF_PREVIEW_SPEC, ShelfPreviewAction} from '../../actions/shelf-preview';
 import {PLOT_HOVER_MIN_DURATION} from '../../constants';
 import {Bookmark} from '../../models/bookmark';
+import {State} from '../../models/index';
 import {PlotFieldInfo, ResultPlot} from '../../models/result';
+import {selectData} from '../../selectors/dataset';
 import {Field} from '../field/index';
 import {Logger} from '../util/util.logger';
 import {VegaLite} from '../vega-lite/index';
 import {BookmarkButton} from './bookmarkbutton';
 import * as styles from './plot.scss';
 
-export interface PlotProps extends ActionHandler<
+export interface PlotOwnProps extends ActionHandler<
   ShelfAction | BookmarkAction | ShelfPreviewAction | ResultAction | LogAction
 > {
   fieldInfos?: PlotFieldInfo[];
@@ -37,6 +41,12 @@ export interface PlotProps extends ActionHandler<
   // so we can close the modal when the specify button is clicked.
   closeModal?: () => void;
 }
+
+export interface PlotConnectProps {
+  data: Data;
+}
+
+export type PlotProps = PlotOwnProps & PlotConnectProps;
 
 export interface PlotState {
   hovered: boolean;
@@ -84,7 +94,7 @@ export class PlotBase extends React.PureComponent<PlotProps, PlotState> {
   }
 
   public render() {
-    const {isPlotListItem, onSort, showBookmarkButton, showSpecifyButton, spec} = this.props;
+    const {isPlotListItem, onSort, showBookmarkButton, showSpecifyButton, spec, data} = this.props;
 
     let notesDiv;
     const specKey = JSON.stringify(spec);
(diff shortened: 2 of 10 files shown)


kanitw:
- `vlSpec.data.name` would give your the data source name
- Use [`view.change`](https://vega.github.io/vega/docs/api/view/#view_change) to update new data when data changes.  (Remove all old, add all new.)

src/actions/result.ts
```diff
@@ -8,7 +8,8 @@ import {State} from '../models/index';
 import {ResultPlot, ResultPlotWithKey} from '../models/result';
 import {ResultType} from '../models/result';
 import {ShelfFieldDef} from '../models/shelf/spec/encoding';
-import {selectConfig, selectData, selectSchema} from '../selectors';
+import {selectConfig, selectSchema} from '../selectors';
```
kanitw: Hmm why do we have to change this here?

src/api/api.ts
```diff
@@ -42,6 +42,16 @@ export function fetchCompassQLRecommend(query: Query, schema: Schema, data: Data
...
       const modelGroup = recommend(query, schema).result;
 
+      // TODO:
+      // - replace this with different cached data source's unique names
+      // once we have multiple cached data source from Leilani's optimizer engine
+      // - (short term) always bind data at runtime
+      if (data) {
+        data = isUrlData(data) ? data : {name: 'source'};
+      } else {
+        data = null as NamedData;
```
kanitw: Hmm how come `data` can be `null`?

src/components/plot/index.tsx
```diff
@@ -132,7 +142,11 @@ export class PlotBase extends React.PureComponent<PlotProps, PlotState> {
           onMouseEnter={this.onMouseEnter}
           onMouseLeave={this.onMouseLeave}
         >
-          <VegaLite spec={spec} logger={this.plotLogger}/>
+          {/*
```
kanitw: I think you already did this, so let's remove this comment :)

src/components/vega-lite/index.tsx
```diff
@@ -11,16 +12,19 @@ export interface VegaLiteProps {
...
 const CHART_REF = 'chart';
 
 export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
+  private view: vega.View;
 
   public render() {
     return (
       <div>
-        <div className='chart' ref={CHART_REF}/>
+        <div className='chart' ref={CHART_REF} style={{width: '200px'}}/>
```
kanitw: Please remove our width hack from our experiement :P

src/components/vega-lite/index.tsx
```diff
@@ -53,13 +57,17 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
...
+      this.view = new vega.View(runtime)
         .logLevel(vega.Warn)
         .initialize(this.refs[CHART_REF] as any)
         .renderer(this.props.renderer || 'svg')
-        .hover()
-        .run();
-      vegaTooltip.vega(view);
+        .hover();
+      vegaTooltip.vega(this.view);
+      if (isInlineData(this.props.data) && isNamedData(vlSpec.data)) {
```
kanitw: We should separate the binding method from the compile method.  
(Suppose `spec` is the same and `data` changes?)

src/components/vega-lite/index.tsx
```diff
@@ -53,28 +57,50 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
...
+    if (prevProps.spec !== this.props.spec) {
+      this.updateSpec(this.props.spec);
+    } else if (prevProps.data !== this.props.data) {
+      this.bindData(this.props.spec);
+    }
+    this.runView();
+  }
+
+  private bindData(vlSpec: TopLevelExtendedSpec) {
+    if (isInlineData(this.props.data) && isNamedData(vlSpec.data)) {
```
kanitw: `vlSpec` is `this.props.spec`, so I don't think we need it to be a param here.

src/components/vega-lite/index.tsx
```diff
@@ -53,28 +57,50 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
...
+      const name = vlSpec.data.name;
+      this.view.change(name, vega.changeset().insert(this.props.data.values).remove(
```
kanitw:
```
this.view.change(name, 
  vega.changeset()
          .insert(data.values) // const {data, spec} = this.props; above
          .remove(() => true)
);
```

src/components/vega-lite/index.tsx
```diff
@@ -53,28 +57,50 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
...
+      this.bindData(vlSpec);
     } catch (err) {
       logger.error(err);
     }
   }
 
   protected componentDidMount() {
-    this.renderVega(this.props.spec);
+    const {spec} = this.props;
+    this.updateSpec(spec);
```
kanitw: What if we already have data when it mounts?

src/components/vega-lite/index.tsx
```diff
@@ -50,31 +55,69 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
...
+  }
+
+  private bindData() {
+    const {data, spec} = this.props;
+    if (isInlineData(data) && isNamedData(spec.data) && !this.view.data(spec.data.name)) {
+      this.view.change(spec.data.name,
+        vega.changeset()
+            .remove(() => true)
+            .insert(data.values)
+           // remove previous data
```
kanitw: I think this comment is misplaced -- I would put it on the same line as `remove(() => true)`

src/components/vega-lite/index.tsx
```diff
@@ -50,31 +55,69 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
...
   }
 
   protected componentDidMount() {
-    this.renderVega(this.props.spec);
+    this.updateSpec();
+    this.runView();
+  }
+
+  protected componentWillUpdate() {
+    this.size = this.getChartSize();
```
kanitw: Can you explain why you have to set size after componentWillUpdate?  

Note that the size of the plot can depend on data, which means that we have to read the size from SVG and set the size to the surrounding container _after_ `view.run()`.

kanitw: Oh, but this is shadow DOM, so maybe you're right.  I'm not sure.  If it works after `view.run()`, I would put it here though.

src/components/vega-lite/index.tsx
```diff
@@ -50,31 +55,69 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
...
 
-  protected componentWillReceiveProps(nextProps: VegaLiteProps) {
-    if (this.props.spec !== nextProps.spec) {
-      setTimeout(() => {
-        this.renderVega(nextProps.spec);
-      });
+  protected componentDidUpdate(prevProps: VegaLiteProps, prevState: {}) {
+    if (prevProps.spec !== this.props.spec) {
+      const chart = this.refs[CHART_REF] as HTMLElement;
+      chart.style.width = this.size[0] + 'px';
```
kanitw: See this [comment thread](https://github.com/vega/voyager/pull/658/files#r135868962).

src/components/vega-lite/index.tsx
starry97: We want to make the plot size persist when clicking the sort button, right? Getting size in componentWillUpdate will give us the previous plot size, and the new plot should be this size.

src/components/vega-lite/index.tsx
```diff
@@ -50,31 +55,69 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, any> {
...
+      chart.style.height = (this.size[1] + 3) + 'px';
+      // add extra 3px to height so chart can contain the svg/canvas plot
```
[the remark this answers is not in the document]
starry97: Because the size the svg is slightly smaller than its container div chart. I inspected the element, and the height of svg is 3px smaller than its div.

PR #659, opened 2017-08-23
Handle errors for load 

starry97:
Part of #588

### DIFF
--- a/src/actions/dataset.ts
+++ b/src/actions/dataset.ts
@@ -57,8 +57,10 @@ export function datasetLoad(name: string, dataset: Data): DatasetLoad {
       const url = dataset.url;
 
       return fetch(url)
-        .then(response => response.json()) // TODO: handle error
-        .then(data => fetchCompassQLBuildSchema(data, config)) // TODO: handle error
+        .then(response => response.json())
+        .catch(errorCatch)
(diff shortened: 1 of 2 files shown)


src/actions/dataset.ts
```diff
@@ -81,7 +84,9 @@ export function datasetLoad(name: string, dataset: Data): DatasetLoad {
     } else {
       throw new Error('dataset load error: dataset type not detected');
     }
-
-
   };
 };
+
+function errorCatch(err: any) {
```
kanitw: Don't Typescript already provide `Error` type or something.  (We should avoid `any`.)

src/components/data-selector/index.tsx
```diff
@@ -162,7 +162,12 @@ export class DataSelectorBase extends React.PureComponent<DataSelectorProps, any
       const name = file.name.replace(/\.\w+$/, '');
       const format = file.name.split('.').pop();
 
-      const values = vega.read(lEvent.target.result, {type: format});
+      let values;
+      try {
+        values = vega.read(lEvent.target.result, {type: format});
+      } catch (err) {
+        window.alert(err);
```
kanitw: Don't this one have to be `err.Message` like the other file?

PR #664, opened 2017-08-24
Add ValueDef support: Model, Reducer, UI

kanitw:
@ZeningQu  -- you can take it from here :)


src/models/shelf/spec/encoding.ts
```diff
@@ -65,10 +65,10 @@ export interface ShelfAnyEncodingDef extends ShelfFieldDef {
   channel: SHORT_WILDCARD;
 }
 
+export type ShelfValueDef = ValueDef;
+
 export type SpecificEncoding = {
-  // TODO: ShelfFieldDef | ValueDef
-  // (Just use ValueDef, no need for special ShelfValueDef)
-  [P in Channel]?: ShelfFieldDef;
+  [P in FieldDefOnlyChannels]?: ShelfFieldDef
```
kanitw: I don't recall having `FieldDefOnlyChannels`. 

You can use `PositionScaleChannel`, `NonPositionScaleChannel`.
You can then manually add `type NonPositionNonScaleChannel = 'text' | 'tooltip' | 'detail' | 'order'`.

src/models/shelf/spec/encoding.ts
```diff
@@ -112,9 +112,8 @@ export function fromValueQuery(encQ: ValueQuery): ShelfValueDef {
   if (isWildcard(encQ.value)) {
     throw Error('Voyager does not support wildcard value');
   }
-  // TODO: return value depending on channel
   return {
-    value: encQ.value
+    value: encQ.value // TODO: when value is undefined
```
kanitw: Why don't you just do it?    It's pretty simple -- just return undefined.  

Note that everytime you push, github will send me an email for this PR since I create it. 
If you do not intend to get feedback, you might want to limit your mini-push and push a series of atomic commits when it's more ready for feedback. ;)

ZeningQu: Ok, will push less often, sorry... Thanks for letting me know :)

ssharif6:
Refer to #816

PR #667, opened 2017-08-28
Loading spinners

### LINKED ISSUE #589: Loading spinners
- [x] Loading Data (the whole app) -- depends  `isLoading` of `Data` ???

- [x] Loading Result (plot list) -- depends on `isLoading` of `Result`

- [ ] Loading Plot -- use local state (this is simply showing spinner before Vega finishes loading)

starry97:
Part of #589 
- [x] Loading Data (the whole app) -- depends isLoading of Data
- [x] Loading Result (plot list) -- depends on isLoading of Result

### DIFF
--- a/src/components/app-root/index.tsx
+++ b/src/components/app-root/index.tsx
@@ -3,10 +3,12 @@ import * as React from 'react';
 import {DragDropContext} from 'react-dnd';
 import HTML5Backend from 'react-dnd-html5-backend';
 import {connect} from 'react-redux';
+import {ClipLoader} from 'react-spinners';
 import * as SplitPane from 'react-split-pane';
-import {Data} from 'vega-lite/build/src/data';
+import {SPINNER_COLOR} from '../../constants';
+import {Dataset} from '../../models/dataset';
 import {State} from '../../models/index';
-import {selectData} from '../../selectors/dataset';
+import {selectDataset} from '../../selectors/dataset';
 import '../app.scss';
 import {DataPane} from '../data-pane/index';
 import {EncodingPane} from '../encoding-pane/index';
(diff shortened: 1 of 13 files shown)


starry97:
image

starry97:
loading

src/components/app-root/index.tsx
```diff
@@ -39,6 +42,7 @@ class AppRootBase extends React.PureComponent<AppRootProps, {}> {
     }
     return (
       <div className="voyager">
+        <ClipLoader color={SPINNER_COLOR} loading={dataset.isLoading}/>
```
kanitw: If users already click the load button, it might be better to hide `<LoadData>` and show only the spinner and a text saying "loading".  (There is no reason what we should keep the "load" button.)

src/components/plot-list/index.tsx
```diff
@@ -31,8 +35,8 @@ export class PlotListBase extends React.PureComponent<PlotListProps, any> {
   }
 
   public render() {
-    const {plots, handleAction, bookmark, limit} = this.props;
-    const plotListItems = plots.slice(0, limit).map((plot, index) => {
+    const {plots, handleAction, bookmark, limit, isLoading} = this.props;
+    const plotListItems = plots ? plots.slice(0, limit).map((plot, index) => {
```
kanitw: You can do `plots && ...components...` (react's preferred way)  instead of `plots ? ...components... : null`

src/components/plot-list/index.tsx
```diff
@@ -47,14 +51,14 @@ export class PlotListBase extends React.PureComponent<PlotListProps, any> {
           bookmark={bookmark}
         />
       );
-    });
-
+    }) : null;
     return (
-      <div>
+      <div style={{position: 'relative'}}>
+        <ClipLoader color={SPINNER_COLOR} loading={isLoading}/>
```
kanitw: Instead of overlaying the spinner with absolute position?  Can't we just show the spinner (with min-height equal to the plot height) when there is no item and show the items otherwise?

src/components/plot-list/index.tsx
```diff
@@ -21,6 +23,8 @@ export interface PlotListProps extends ActionHandler<ShelfAction|ResultAction> {
   bookmark: Bookmark;
 
   limit?: number;
+
+  isLoading: boolean;
```
kanitw: PlotList takes almost everything in Result as parameters (except query): `plots`, `limit`, `isLoading`.
Maybe we should just take `result` as the parameter instead?

src/components/app.scss
```diff
@@ -151,3 +151,7 @@
     max-width: 200px;
     text-align: left;
   }
+
+:global .react-spinners--clip {
```
kanitw: Can we not add a CSS style to the global scope?

(I guess this probably works if it's nested inside `.voyager` too.)

Issue #623, opened 2017-08-17
Refactor filterReducer to only take filter and action as input


kanitw:
- [x] Instead of taking in the whole shelf spec, we don't need other part of the spec.
- [x] It seems like we only need `schema` for one of them, so maybe moving the logic for getting domain to filter in the models so we can still test, but simplify the reducer to take only `(filter, action)` as input. 

- [x] We can similarly refactor `schemaReducer` (Instead of taking in the whole dataset spec, we don't need other part of the spec.)

kanitw:
- [ ] We can similarly refactor `schemaReducer`

PR #669, opened 2017-08-29
Refactor filterReducer to only take filter and action as input	

starry97:
Part of #623

### DIFF
--- a/src/reducers/shelf/filter.test.ts
+++ b/src/reducers/shelf/filter.test.ts
@@ -1,111 +1,82 @@
 
-import {Schema} from 'compassql/build/src/schema';
 import {OneOfFilter, RangeFilter} from 'vega-lite/build/src/filter';
 import {TimeUnit} from 'vega-lite/build/src/timeunit';
 import {
   FILTER_ADD, FILTER_CLEAR, FILTER_MODIFY_EXTENT, FILTER_MODIFY_MAX_BOUND,
   FILTER_MODIFY_MIN_BOUND, FILTER_MODIFY_ONE_OF, FILTER_MODIFY_TIME_UNIT, FILTER_REMOVE
 } from '../../actions';
-import {DEFAULT_SHELF_UNIT_SPEC} from '../../models/shelf';
-import {ShelfUnitSpec} from '../../models/shelf/spec';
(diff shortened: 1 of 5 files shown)


src/models/shelf/filter.ts
```diff
@@ -114,3 +114,31 @@ export function containsFilter(filters: Array<RangeFilter | OneOfFilter>, target
   }
   return false;
 }
+
+export function getModifyTimeUnitFunction(timeUnit: TimeUnit, domain: number[]) {
+  if (!timeUnit) {
+    return (filter: RangeFilter) => {
```
kanitw: This line isn't tested?

src/reducers/shelf/filter.test.ts
```diff
@@ -1,175 +1,147 @@
...
 import {
   FILTER_ADD, FILTER_CLEAR, FILTER_MODIFY_EXTENT, FILTER_MODIFY_MAX_BOUND,
   FILTER_MODIFY_MIN_BOUND, FILTER_MODIFY_ONE_OF, FILTER_MODIFY_TIME_UNIT, FILTER_REMOVE
 } from '../../actions';
-import {DEFAULT_SHELF_UNIT_SPEC} from '../../models/shelf';
-import {ShelfUnitSpec} from '../../models/shelf/spec';
 import {filterReducer} from './filter';
 
-const rangeFilter: RangeFilter = {field: 'q1', range: [1437978615, 1501137015]};
+const range = [1437978615, 1501137015];
```
kanitw: It seems like most of these constants are used only once.   If that's the case, it's better to just put them inline where there test are.  Putting them up here separate them from the code, making it hard to reader and understand the tests.

(We should only put reused stuff up here.)

src/models/shelf/filter.ts
```diff
@@ -114,3 +114,31 @@ export function containsFilter(filters: Array<RangeFilter | OneOfFilter>, target
   }
   return false;
 }
+
+export function getModifyTimeUnitFunction(timeUnit: TimeUnit, domain: number[]) {
```
kanitw: This logic is only called in reducer so it could just be in the reducer.
(I suggest model only if it is called directly in component.)

PR #674, opened 2017-08-30
Do not output inline data since it makes the log file too big

### LINKED ISSUE #574: Log data too big
Handle the "load" action

starry97:
Fix  #574

### DIFF
--- a/src/models/export-logs.ts
+++ b/src/models/export-logs.ts
@@ -1,13 +1,25 @@
+import {isInlineData} from 'vega-lite/build/src/data';
+import {DATASET_RECEIVE} from '../actions/dataset';
+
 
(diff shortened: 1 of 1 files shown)


src/models/export-logs.ts
```diff
@@ -1,13 +1,24 @@
...
+
 
 export function constructLogString(inputLogs: Array<{action: {type: string, payload: any}, timestamp: number}>):
   Array<{type: string, timestamp: number, ISOString: string, payload: string}> {
   const outputLogs = [];
   for (const inputLog of inputLogs) {
+    const type = inputLog.action.type;
+    let payload: any = inputLog.action.payload;
+    if (type === DATASET_RECEIVE && isInlineData(payload.data)) {
+      payload = {
```
kanitw: Add comment explaining why we don't use inline data.

Issue #673, opened 2017-08-30
Request for trimming size of build for lib-voyager


ajainarayanan:
Thank you so much for re-writing voyager in react and making it open source! Super awesome thing to do!

- [x] Check for duplicate issues
  - Couldn't find any issue that talks about build size for lib voyager for embedding in other projects.
- [x] Describe how to reproduce the bug / the goal of the new feature request
  - **Library Build size**

Right now if I install `datavoyager` from npm it installs the entire project and the `voyager-lib.js` file is 28mb in size. 

I believe the reason for this size is it comes with all the dependencies pre-bundled + unminified. 

I am creating this issue to request to mark certain projects as "external" dependencies in `webpack.config.lib.js` and use webpack's `UglifyJS` to uglify `voyager-lib.js`. For instance, a person who is going to use this are predominantly going  be having `react` and `react-dom` installed. If we could externalize some projects (react based projects) then it would be super helpful to embed datavoyager as a component in other projects.

  - [ ] The other request is to remove the header and footer in the library build as anybody who is going to embed voyager would not require a header and a footer to be attached. 
 
  - [ ] Separate out the css between common and component specific styles since the existing css file modifies nodes like `html, body, h<>tags` that will cause problems in the environment where it will be embedded.

If the contributors of the project agree about externalizing some projects and trimming down library build I would be more than happy to create a PR. This does not change the functionality. Its a change in the build process and the top level app component (exclude header and footer)

kanitw:
Thanks for asking. 

> I am creating this issue to request to mark certain projects as "external" dependencies in `webpack.config.lib.js` and use webpack's `UglifyJS` to uglify `voyager-lib.js`. For instance, a person who is going to use this are predominantly going  be having `react` and `react-dom` installed. If we could externalize some projects (react based projects) then it would be super helpful to embed datavoyager as a component in other projects.

- I think uglifying voyager-lib.js and extracting external dependencies totally make sense. Feel free to submit a PR.  If you want to do so, please also submit a PR to modify [`voyager-electron`](https://github.com/vega/voyager-electron) to make sure it still works with the new lib file. 

> the header and footer in the library build as anybody who is going to embed voyager would not require a header and a footer to be attached. 

[...]

PR #678, opened 2017-08-31
Reduce build size for voyager lib

ajainarayanan:
Fixes:  https://github.com/vega/voyager/issues/673
#### Before build size:
`
 lib-voyager.js    20.2 MB       0  [emitted]  [big]  bundle
`
#### After changes build size:
`
lib-voyager.js    1.04 MB       0  [emitted]  [big]  bundle
`
- Adds following packages as external during voyager library build step,
```
  - react-dom
  - react-css-modules
  - react-dnd
  - react-dnd-html5-backend
  - react-redux
  - react-tether
  - redux
  - redux-thunk
  - redux-undo
  - vega
  - vega-lite
  - vega-tooltip
  - moment
  - font-awesome
  - font-awesome-sass-loade
```

- Adds separate webpack config for electron build to be used in a electron based project
- Fixes test script to exclude externals during unit test build.
- Verified lint & unit tests. All passed successfully.

#### Note:
- Yet to bump version in package.json


config/webpack.config.lib.js
```diff
@@ -27,11 +27,6 @@ module.exports = {
     libraryTarget: 'umd',
   },
 
-  // Enable sourcemaps for debugging webpack's output.
-  // If it is inline, it will break CSS sourcemaps because of
-  // `extract-text-webpack-plugin`
-  devtool: 'source-map',
```
kanitw: I think having totally no sourcemap at all will make it harder to users lib-voyager to report bug?

config/webpack.config.lib.js
```diff
@@ -108,14 +100,37 @@ module.exports = {
...
     new webpack.DefinePlugin(env.stringified),
 
+    new webpack.optimize.UglifyJsPlugin({
+      compress: {
+        warnings: false,
+      },
+      output: {
+        comments: false,
+      },
+      sourceMap: false,
```
kanitw: Maybe better to include source map?

config/webpack.config.electron.lib.js
```diff
@@ -0,0 +1,18 @@
+var path = require('path');
```
kanitw: Can you explain why we need separate webpack  config for electron build? 

Ideally, if we can make it work without adding one more type of build, that will be simpler as the current build setup is already complicate enough.

scripts/jest-ui.js
```diff
@@ -16,6 +16,7 @@ const glob = require('glob');
 config.output.publicPath = '/lib-test/';
 config.output.path = path.resolve(__dirname, '../lib-test');
 config.output.filename = '[name].js';
+config.externals = {};
```
kanitw: Can you explain more what is this line for?

package.json
```diff
@@ -24,7 +24,8 @@
     "start:server": "cross-env SERVER=\"http://localhost:3000\" webpack-dev-server --open --config config/webpack.config.dev.js",
     "build": "node scripts/build.js",
     "postbuild": "npm run tsc && npm run build:lib",
-    "build:lib": "webpack -d --config config/webpack.config.lib.js",
+    "build:lib": "webpack --config config/webpack.config.lib.js",
+    "build:electron": "webpack --config config/webpack.config.electron.lib.js",
```
kanitw: As mentioned in another comment, please explain why we need another config.

kanitw:
Thanks for your PR.  I have some comments.

ajainarayanan:
@kanitw 
Regarding sourcemaps, on a production environment libraries generally don't add source maps. This is to reduce the build size further down. If you feel strongly about it I can add it back.

Regarding electron build, since we have marked a few modules as external the build output of webpack is no longer the same for web vs electron environments. The separate config is for building voyager for electron based environment. For reference these are the targets that webpack can build to: https://webpack.js.org/configuration/target/

scripts/jest-ui.js
ajainarayanan: During test we still use `webpack.config.lib.js` but since I have added `externals` there during test build webpack again excludes those modules out which will lead to tests failing because it is not able to find some modules. This is to make sure the test build is till the entire build without any excluded modules.

ajainarayanan:
@kanitw A nice thing to have is to limit the files during npm publish. If during `npm publish` if you can restrict only the build folder to be published it would be nicer as `npm i datavoyager` would just download the build folder instead of the entire project.

config/webpack.config.lib.js
```diff
@@ -46,9 +46,6 @@ module.exports = {
   module: {
     rules: [
       { test: /\.tsx?$/, use: "ts-loader" },
-      // All output '.js' files will have any sourcemaps re-processed by 'source-map-loader'.
```
kanitw: Could you explain why do you need to remove `source-map-loader` here?

kanitw:
> @kanitw A nice thing to have is to limit the files during npm publish. If during npm publish if you can restrict only the build folder to be published it would be nicer as npm i datavoyager would just download the build folder instead of the entire project.

Sounds good.  Feel free to submit another PR to support this behavior.  :)

config/webpack.config.lib.js
ajainarayanan: Sorry didn't add this while adding back source maps. my bad.

ajainarayanan:
@kanitw Have updated the PR. Let me know if I need to squash the commits. Thanks for the review!

Issue #617, opened 2017-08-15
Extract Filtering from spec


kanitw:
- [x] Loading filter should load filter to the right part
  - move `SPEC_LOAD` to `reducers/shelf/index`
- [x] apply filter separately using dataflow API

- [x] Stick in the right filter transforms when we 
  - [x] bookmark 
  - [x] copy

- [x] timeunit

Test
- [x] if this makes filter smoother (it shouldn't update the whole query set).

kanitw:
Done in https://github.com/vega/voyager/pull/697

PR #683, opened 2017-09-01
Move `filters` to be a part of `Shelf`, not `ShelfUnitSpec`	

kanitw:
Part of https://github.com/vega/voyager/issues/617


src/models/shelf/filter.ts
```diff
@@ -1,10 +1,34 @@
...
 import {ShelfFieldDef} from './spec';
 
+export type ShelfFilter = RangeFilter | OneOfFilter;
+
+export function fromTransforms(transforms: Transform[]): ShelfFilter[] {
+  if (!transforms) {
+    return [];
+  } else {
+    const filters: Array<RangeFilter|OneOfFilter> = [];
+    transforms.map(transform => {
```
starry97: I thought you said we should change this to foreach

PR #690, opened 2017-09-02
Customize scale type

starry97:


src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -40,14 +42,22 @@ interface EncodingShelfProps extends EncodingShelfPropsBase, EncodingShelfDropTa
...
+    const {id, connectDropTarget, fieldDef, handleAction} = this.props;
 
     const isWildcardShelf = isWildcard(id.channel);
     const channelName = isWildcardShelf ? 'any' : id.channel;
 
     return connectDropTarget(
       <div styleName={isWildcardShelf ? 'wildcard-shelf' : 'encoding-shelf'}>
-        <div styleName="shelf-label">{channelName}</div>
+        <div styleName="shelf-label">
+          <FieldCustomizer
```
kanitw: I think caret and tether related stuff should be a part of shelf. 

Basically, you can imagine in the future the EncodingShelf's caret open either a FieldCustomizer or ValueCustomizer depending on whether the channelDef is a fieldDef or not.   (Right now we of course only support fieldDef.)

kanitw:
- [ ] The popup commands are so big.  I'd make the font style match existing items.  
- [ ] Placing caret on the left while most of them are empty makes visual style weird.  Maybe we can try placing them on the right?
- [ ] As mentioned earlier, one line layout would work better

**Scale**
Type         [     log         ]

- [ ] I don't think we need submit button.  It unnecessary increases the number of clicks we need to do one action.  

## In a follow up PR

- [ ] We need to exclude unsupported scale type but that requires latest version of Vega-Lite.

kanitw:
- [ ] Click outside to disable.

src/components/encoding-pane/property-editor.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public render() {
+    const {prop, nestedProp, fieldDef} = this.props;
+    const uiSchema = {
+      "ui:title": `${prop} ${nestedProp}`,
+      "ui:placeholder": "auto",
+      "ui:emptyValue": "auto"
+    };
+    const formaData = fieldDef.scale ? fieldDef.scale.type : 'auto';
```
kanitw: formData?

PR #692, opened 2017-09-02
Add optional config when embedding voyager as library

ajainarayanan:
- Add default embed configuration to `VoyagerConfig`.
- Currently it supports hiding header and footer.
- This configuration basically gives an option to not render the header and footer when embedding voyager in another application.
- By default the header and footer are rendered. So existing setup requires no change.

Created this as part of the discussion that went in this Issue: https://github.com/vega/voyager/issues/673

#### Note:
- Have not modified any docs. Not sure if this is the final implementation or if there is a simpler way to do this.

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -132,9 +132,9 @@ module.exports = {
   // assume a corresponding global variable exists and use that instead.
   // This is important because it allows us to avoid bundling all of our
   // dependencies, which allows browsers to cache those libraries between builds.
+
+  // Since its a local dev build we don't need to mark anything external.
   externals: {
-    "react": "React",
-    "react-dom": "ReactDOM"
   },
   plugins: [
     new webpack.HotModuleReplacementPlugin(),
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -133,8 +133,9 @@ module.exports = {
   // This is important because it allows us to avoid bundling all of our
   // dependencies, which allows browsers to cache those libraries between builds.
 
-  // Since its a local dev build we don't need to mark anything external.
   externals: {
+    "react": "React",
+    "react-dom": "ReactDOM"
   },
   plugins: [
     new webpack.HotModuleReplacementPlugin(),
--- a/src/components/app-root/index.tsx
+++ b/src/components/app-root/index.tsx
@@ -6,8 +6,10 @@ import {connect} from 'react-redux';
 import {ClipLoader} from 'react-spinners';
 import * as SplitPane from 'react-split-pane';
 import {SPINNER_COLOR} from '../../constants';
+import {VoyagerConfig} from '../../models/config';
 import {Dataset} from '../../models/dataset';
 import {State} from '../../models/index';
+import {selectConfig} from '../../selectors';
 import {selectDataset} from '../../selectors/dataset';
 import '../app.scss';
 import {DataPane} from '../data-pane/index';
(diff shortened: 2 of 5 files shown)


ajainarayanan:
@kanitw As discussed in #673 I have added an optional configuration to show/hide header and footer. Let me know if I have missed anything.

config/webpack.config.dev.js
```diff
@@ -132,9 +132,9 @@ module.exports = {
   // assume a corresponding global variable exists and use that instead.
   // This is important because it allows us to avoid bundling all of our
   // dependencies, which allows browsers to cache those libraries between builds.
+
+  // Since its a local dev build we don't need to mark anything external.
```
kanitw: Like commented above, not bundling all of our dependencies allow browsers to cache those libraries and may make the bundling faster?

src/components/app-root/index.tsx
```diff
@@ -39,13 +45,15 @@ class AppRootBase extends React.PureComponent<AppRootProps, {}> {
...
+        if (!hideFooter) {
+          footer = <Footer/>;
+        }
       }
     }
     return (
       <div className="voyager">
         <LogPane/>
-        <Header/>
+        {!hideHeader ? <Header/> : null}
```
kanitw: `!hideHeader && <Header/>`

src/lib-voyager.tsx
```diff
@@ -44,7 +44,7 @@ export class Voyager {
       this.container = container;
     }
 
-    this.config = config;
+    this.config = Object.assign({}, DEFAULT_VOYAGER_CONFIG, config);
```
kanitw: `{...DEFAULT_VOYAGER_CONFIG, ...config}`

(We use the spread operator in this project.)

src/models/config.ts
```diff
@@ -1,11 +1,20 @@
+export interface EmbedProps {
+  hideHeader?: boolean;
+  hideFooter?: boolean;
+};
 export interface VoyagerConfig {
   showDataSourceSelector?: boolean;
   serverUrl?: string | null;
   manualSpecificationOnly?: boolean;
+  embedProps?: EmbedProps;
```
kanitw: Let's not include unnecessary nesting.  Most of props are for using Voyager as a library anyway.  

Just make `hideHeader/Footer` properties of VoyagerConfig.

config/webpack.config.dev.js
ajainarayanan: Sorry didn't mean to overwrite. Since we modified external thought this might be an issue. Have reverted back.

ajainarayanan:
@kanitw Have updated the PR.

src/components/app-root/index.tsx
```diff
@@ -20,11 +22,13 @@ import {ViewPane} from '../view-pane/index';
...
 export interface AppRootProps {
   dataset: Dataset;
+  config: VoyagerConfig;
 }
 
 class AppRootBase extends React.PureComponent<AppRootProps, {}> {
   public render() {
-    const {dataset} = this.props;
+    const {dataset, config} = this.props;
+    const {hideHeader = false, hideFooter = false} = config;
```
kanitw: Given that we already set default somewhere else, maybe the default parameter here is unnecessary?

ajainarayanan: Its just as an extra precaution. I guess I really need to get used to typescript :) If its un-necessary I can remove this.

ajainarayanan:
@kanitw Have removed the defaults as they are provided by the config.

ajainarayanan:
@kanitw Thanks! Are there any tentative time line when you would cut a release

kanitw:
We plan to cut another release sometimes this week.

PR #698, opened 2017-09-05
Add loading spinner for vega lite plot

### LINKED ISSUE #589: Loading spinners

starry97:
Fixed #589
loading

### DIFF
--- a/src/components/vega-lite/index.tsx
+++ b/src/components/vega-lite/index.tsx
@@ -1,9 +1,11 @@
 import * as React from 'react';
+import {ClipLoader} from 'react-spinners';
 import * as vega from 'vega';
 import * as vl from 'vega-lite';
 import {Data, isInlineData, isNamedData} from 'vega-lite/build/src/data';
 import {TopLevelExtendedSpec} from 'vega-lite/build/src/spec';
 import * as vegaTooltip from 'vega-tooltip';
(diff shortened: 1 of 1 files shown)


src/components/vega-lite/index.tsx
```diff
@@ -80,8 +94,14 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, {}> {
       clearTimeout(this.mountTimeout);
     }
     this.mountTimeout = setTimeout(() => {
+      this.setState({
```
kanitw: Shouldn't we set isLoading to `true` outside `setTimeout`?

(It's suppose to be pretty fast, and just in case other things block us this timeout to execute we still have isLoading ready before hand.

src/components/vega-lite/index.tsx
```diff
@@ -94,6 +114,9 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, {}> {
       clearTimeout(this.updateTimeout);
     }
     this.updateTimeout = setTimeout(() => {
+      this.setState({
```
kanitw: same here

Issue #702, opened 2017-09-06
Use production build of React


kanitw:
We are now receiving a warning in our UI test

```
 console.error lib-test/components/app.test.ui.tsx.js:1
      Warning: It looks like you're using a minified copy of the development build of React. When deploying React apps to production, make sure to use the production build which skips development warnings and is faster. See ht
tps://fb.me/react-minification for more details.
    console.error lib-test/components/app.test.ui.tsx.js:1
      You are currently using minified code outside of NODE_ENV === 'production'. This means that you are running a slower development build of Redux. You can use loose-envify (https://github.com/zertosh/loose-envify) for brow
serify or DefinePlugin for webpack (http://stackoverflow.com/questions/30030031) to ensure you have the correct code for your production build.
```

kanitw:
@ajainarayanan -- this might be related to the recent changes in https://github.com/vega/voyager/blob/master/config/webpack.config.lib.js?

ajainarayanan:
@kanitw Yes. I believe so. Have created a fix pull request for it. Can you help review it.

PR #703, opened 2017-09-06
Avoids uglifying library when running test build #702

ajainarayanan:
#### Issue:
- React warns users when a minified build of react is run in a non-production environment
- Recently we started using UgilifyJS webpack plugin to uglify final library build. However since we used the same webpack config it was minifying development version of react which was causing the issue.(UI test build runs with `NODE_ENV === 'test'` and react expects this to be `production` if minified).

#### Fix:
- Avoids uglifying code when running ui test build.

Fixes https://github.com/vega/voyager/issues/702


kanitw:
Thanks for your PR!

PR #705, opened 2017-09-06
Change filter button to toggle button

### LINKED ISSUE #696: Clicking the filter button for the second time does not disable it.
Instead it shows error that the same filter can't be added twice.  

We also may want to distinguish the toggle state for the filter button (maybe show the icon faded if filter is not activated yet).

starry97:
Fix #696


starry97:
image

PR #706, opened 2017-09-08
Add tether component wrapper so clicking outside can close the attached popup

starry97:


kanitw:
Clicking outside doesn't seem to work. 

Plus, I think the component is not cleanly decoupled from other components.   (The state is still in field/bookmarkbutton, etc.)  Thus, this refactor is probably not worth it.

Issue #555, opened 2017-08-09
Custom Wildcard Field


kanitw:
- [ ] Drag and drop to add field.
- [ ] Pop-up

PR #709, opened 2017-09-11
Add custom-wildcard-field actions, reducers, and selectors

felixcodes:
Part of #555

### DIFF
--- a/src/models/custom-wildcard-field.ts
+++ b/src/models/custom-wildcard-field.ts
@@ -1,0 +1,9 @@
+import {ExpandedType} from 'compassql/build/src/query/expandedtype';
+
+export interface CustomWildcardField {
+  fields: string[];
+  wildcardDescription: string;
+  type: ExpandedType;
+}
+
+export const DEFAULT_CUSTOM_WILDCARD_FIELDS: CustomWildcardField[] = [];
--- a/src/reducers/custom-wildcard-field.ts
+++ b/src/reducers/custom-wildcard-field.ts
@@ -1,0 +1,80 @@
+
+import {toMap} from 'compassql/build/src/util';
+import {Action} from '../actions';
+import {CUSTOM_WILDCARD_ADD, CUSTOM_WILDCARD_ADD_FIELD, CUSTOM_WILDCARD_MODIFY_WILDCARD_DESCRIPTION,
+        CUSTOM_WILDCARD_REMOVE, CUSTOM_WILDCARD_REMOVE_FIELD} from '../actions/custom-wildcard-field';
+import {CustomWildcardField} from '../models/custom-wildcard-field';
+import {insertItemToArray, modifyItemInArray, removeItemFromArray} from './util';
+
+function modifyFieldsProperty(fields: string[]) {
+  return (customWildcardField: CustomWildcardField) => {
+    return {
+      ...customWildcardField,
+      fields
+    };
+  };
+}
+
+export function customWildcardFieldReducer(
+  customWildcardFields: Readonly<CustomWildcardField[]> = [],
+  action: Action
+): CustomWildcardField[] {
+  switch (action.type) {
+    case CUSTOM_WILDCARD_ADD: {
+      const {fields, type} = action.payload;
+
+      let index = action.payload.index;
+      if (!index) {
+        index = customWildcardFields.length;
+      }
+
+      return insertItemToArray(customWildcardFields, index, {
+        fields,
+        type,
+        wildcardDescription: null
+      });
+    }
+
+    case CUSTOM_WILDCARD_REMOVE: {
+      const {index} = action.payload;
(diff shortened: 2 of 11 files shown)


src/reducers/index.test.ts
```diff
@@ -34,7 +35,7 @@ describe('reducers/index', () => {
   });
 
   describe('RESET', () => {
-    it('should reset bookmark, dataset, shelf, result', () => {
+    it('should reset bookmark, dataset, shelf, result, customWildcards', () => {
```
felixcodes: customWildcardFields**

src/reducers/custom-wildcard-field.ts
```diff
@@ -0,0 +1,92 @@
...
+    case CUSTOM_WILDCARD_ADD_FIELD: {
+      const {index, fields} = action.payload;
+
+      const originalFields: string[] = customWildcardFields[index].fields;
+      const originalFieldsIndex = originalFields.reduce((dict, field) => {
+        dict[field] = true;
+        return dict;
+      }, {});
+
+      let updatedFields: string[];
```
kanitw: - I would name this `addedFields`

- I don't understand why you need to separate the logic for `n>1` and `n=1`.   Using a functional style, you can write this in one-line-ish. 

- There is another problem with this logic for immutability at the end.  What is it?  (Think for another 5 mins and ask me if you don't know.)

src/reducers/custom-wildcard-field.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      return modifyItemInArray(customWildcardFields, index, modifyFieldsProperty(updatedFields));
+    }
+
+    case CUSTOM_WILDCARD_REMOVE_FIELD: {
+      const {index, field} = action.payload;
+      const originalFields: string[] = customWildcardFields[index].fields;
+      const updatedFields = originalFields.filter(originalField => originalField !== field);
+
+      return modifyItemInArray(customWildcardFields, index, modifyFieldsProperty(updatedFields));
```
kanitw: Similar problem with immutability here.

src/reducers/index.test.ts
```diff
@@ -45,6 +46,9 @@ describe('reducers/index', () => {
           ...DEFAULT_UNDOABLE_STATE,
           present: {
             ...DEFAULT_UNDOABLE_STATE_BASE,
+            customWildcardFields: {
```
kanitw: Why do you need to add this?

src/selectors/shelf.test.ts
```diff
@@ -16,6 +17,7 @@ describe('selectors/shelf', () => {
         undoable: {
           ...DEFAULT_STATE.undoable,
           present: {
+            customWildcardFields: DEFAULT_CUSTOM_WILDCARD_FIELDS,
```
kanitw: Instead of adding this, add `...DEFAULT_UNDOABLE_STATE_BASE` and remove `result: DEFAULT_RESULT_INDEX`

src/reducers/custom-wildcard-field.ts
felixcodes: Need to fix the n=0 corner case (add a duplicate field -- nothing changes, don't return a new array)

src/reducers/custom-wildcard-field.ts
felixcodes: N/A - a remove action is triggered because the user deletes an existing field -- array.filter() will always return a new array

src/reducers/custom-wildcard-field.ts
```diff
@@ -0,0 +1,82 @@
...
+        return {
+          ...customWildcardField,
+          title
+        };
+      };
+      return modifyItemInArray(customWildcardFields, index, modifyTitle);
+    }
+
+    default: {
+      return customWildcardFields;
```
kanitw: (minor) -- there is no need for `default:` and braces,  just return this at the end.

src/models/custom-wildcard-field.ts
```diff
@@ -0,0 +1,9 @@
+import {ExpandedType} from 'compassql/build/src/query/expandedtype';
+
+export interface CustomWildcardField {
+  fields: string[];
```
kanitw: This won't be compatible with how ShelfFieldDef is structured. 

This should be `field: Wildcard<string>` so you can drag and drop wildcard field and it works right away.

src/models/custom-wildcard-field.ts
```diff
@@ -0,0 +1,9 @@
...
+  title: string;
```
kanitw: Rename this to `wildcardDescription`

src/models/custom-wildcard-field.ts
kanitw: Let's punt on this to see if it's a good idea in the next PR.

Issue #721, opened 2017-09-14
Ability to customize vega plots rendered


ajainarayanan:
#### New feature request.
- Request to add the ability to customize chart (fill color, padding etc.,)

#### Existing feature:
 - We already have `setSpec` public API but that requires setting the entire spec for each sections of voyager. 

#### Approach:
 - Allow `VoyagerConfig` to have an object like `vegaPlotSpec` or `vegaPlotStyle` (not strong about name) that will allow users to provide `config` portion of the spec ([Configuration](https://vega.github.io/vega-lite/docs/config.html))

For eg.,

```
var voyagerInstance = CreateVoyager(myContainer, {
  "config": {
    "mark": {
      "color": "black"
    }
  }
}, data);
```
Or 
```
voyagerInstance.setApplicationState({
  ...voyagerState,
  config: {
    ...voyagerState.config,
    vegaPlotSpec: {
      "config": {
        "mark": {
          "color": "black"
        }
      }
    }
});
```

If contributors agree on the idea and the approach I can create a PR. Just talking out loud to see if it is the correct approach or if there is an easier way to do this.

kanitw:
I think it is good have have such support. 

Vega-Lite `config` is already a part of our redux store model  (under `ShelfUnitSpec`). 

Here are the steps that I have in mind:

- [ ] Add `setVegaLiteConfig` to the `Voyager` class API and the `App` Component  
  - [ ] You can add `vegaliteConfig` as a property of `Voyager` class and a prop of the `App` component (just like how we have `config`) 

You will also have to modify the reducer
- [ ] Add a new action `SPEC_CONFIG_CHANGE` to the `shelfSpecReducer` so that the `App`'s `setVegaLiteConfig` can dispatch the action
- [ ] Modify `SPEC_CLEAR` action to preserve the stored config 
- [ ] Make sure `SPEC_LOAD` and `SHELF_LOAD_QUERY` preserve stored `spec.config` -- (I think it should already be the case, but we may not have unit test for it -- so write one for each to ensure). 

If I overlook anything / if you disagree with any part, please feel free to let me know!

ajainarayanan:
@kanitw Ok will play around with it. Looks like I missed the vega lite config in redux store. Will try to see if I can create a PR this weekend. 🤞

kanitw:
Sounds good and don't worry. It's quite a huge store.  I am glad you asked and am happy to help! :)

ajainarayanan:
[...]

PR #726, opened 2017-09-18
Ability to customize vega-lite charts via configuration

ajainarayanan:
- Adds the ability to pass on config to customize vegalite graphs.
- References: https://github.com/vega/voyager/issues/721 

#### Note:
- Have not modified README about the new API. We can update once the implementation is reviewed.

### DIFF
--- a/src/components/plot/index.tsx
+++ b/src/components/plot/index.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import * as CopyToClipboard from 'react-copy-to-clipboard';
 import * as CSSModules from 'react-css-modules';
+import {connect} from 'react-redux';
 import * as TetherComponent from 'react-tether';
 import {InlineData} from 'vega-lite/build/src/data';
 import {isDiscrete, isFieldDef} from 'vega-lite/build/src/fielddef';
@@ -14,15 +15,18 @@ import {ShelfAction, SPEC_LOAD} from '../../actions/shelf';
 import {SHELF_PREVIEW_DISABLE, SHELF_PREVIEW_SPEC, ShelfPreviewAction} from '../../actions/shelf-preview';
 import {PLOT_HOVER_MIN_DURATION} from '../../constants';
 import {Bookmark} from '../../models/bookmark';
+import { VoyagerConfig } from '../../models/config';
+import {State} from '../../models/index';
 import {PlotFieldInfo, ResultPlot} from '../../models/result';
 import {ShelfFilter, toTransforms} from '../../models/shelf/filter';
+import {selectConfig} from '../../selectors';
 import {Field} from '../field/index';
 import {Logger} from '../util/util.logger';
 import {VegaLite} from '../vega-lite/index';
 import {BookmarkButton} from './bookmarkbutton';
 import * as styles from './plot.scss';
 
-export interface PlotProps extends ActionHandler<
+export interface PlotOwnProps extends ActionHandler<
   ShelfAction | BookmarkAction | ShelfPreviewAction | ResultAction | LogAction
 > {
   data: InlineData;
@@ -42,6 +46,11 @@ export interface PlotProps extends ActionHandler<
   closeModal?: () => void;
 }
 
+export interface PlotConnectProps {
+  voyagerConfig: VoyagerConfig;
+}
+
+export type PlotProps = PlotOwnProps & PlotConnectProps;
 
 export interface PlotState {
   hovered: boolean;
@@ -89,7 +98,7 @@ export class PlotBase extends React.PureComponent<PlotProps, PlotState> {
   }
 
   public render() {
-    const {isPlotListItem, onSort, showBookmarkButton, showSpecifyButton, spec, data} = this.props;
+    const {isPlotListItem, onSort, showBookmarkButton, showSpecifyButton, spec, data, voyagerConfig} = this.props;
 
     let notesDiv;
     const specKey = JSON.stringify(spec);
@@ -137,7 +146,7 @@ export class PlotBase extends React.PureComponent<PlotProps, PlotState> {
           onMouseEnter={this.onMouseEnter}
           onMouseLeave={this.onMouseLeave}
         >
-          <VegaLite spec={spec} logger={this.plotLogger} data={data}/>
+          <VegaLite spec={spec} logger={this.plotLogger} data={data} vegaliteConfig={voyagerConfig.vegaliteConfig}/>
         </div>
         {notesDiv}
       </div>
@@ -341,4 +350,11 @@ export class PlotBase extends React.PureComponent<PlotProps, PlotState> {
   }
 }
 
-export const Plot = CSSModules(PlotBase, styles);
+
+export const Plot = connect<PlotConnectProps, {}, PlotOwnProps>(
+  (state: State) => {
+    return {
+      voyagerConfig: selectConfig(state)
+    };
+  }
+)(CSSModules(PlotBase, styles));
--- a/src/components/vega-lite/index.tsx
+++ b/src/components/vega-lite/index.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {ClipLoader} from 'react-spinners';
 import * as vega from 'vega';
 import * as vl from 'vega-lite';
+import {Config} from 'vega-lite/build/src/config';
 import {InlineData, isNamedData} from 'vega-lite/build/src/data';
 import {TopLevelExtendedSpec} from 'vega-lite/build/src/spec';
 import * as vegaTooltip from 'vega-tooltip';
@@ -16,6 +17,8 @@ export interface VegaLiteProps {
   logger: Logger;
 
   data: InlineData;
+
+  vegaliteConfig: Config;
 }
 
 export interface VegaLiteState {
@@ -72,8 +75,15 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, VegaLiteState> 
     //     "y": {"field": "b", "type": "quantitative"}
     //   }
     // };
-    const {logger} = this.props;
-    const vlSpec = this.props.spec;
+    const {logger, vegaliteConfig} = this.props;
+    let vlSpec = this.props.spec;
+    vlSpec = {
+      ...vlSpec,
+      config: {
+        ...vlSpec.config,
+        ...vegaliteConfig
+      }
+    };
     try {
       const spec = vl.compile(vlSpec, logger).spec;
       const runtime = vega.parse(spec, vlSpec.config);
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -4,6 +4,7 @@ import { Provider } from 'react-redux';
 
 import 'font-awesome-sass-loader';
 
+import {Config} from 'vega-lite/build/src/config';
 import { Data } from 'vega-lite/build/src/data';
 import {App} from './components/app';
 import { VOYAGER_CONFIG } from './constants';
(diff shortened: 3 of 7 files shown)


ajainarayanan:
@kanitw Have tried what we discussed. Let me know if I have missed something.

- Ran through dev build
- Ran test
- Tested lib-voyager in a different project

All seemed to work the way it is expected.

src/components/plot/index.tsx
```diff
@@ -341,4 +350,11 @@ export class PlotBase extends React.PureComponent<PlotProps, PlotState> {
   }
 }
 
-export const Plot = CSSModules(PlotBase, styles);
+
+export const Plot = connect<PlotConnectProps, {}, PlotOwnProps>(
+  (state: State) => {
+    return {
+      voyagerConfig: selectConfig(state)
```
kanitw: Plot should be a dumb (renderer-only) component, so it should not use `connect`.   
Instead, `config` should be passed from its parent component, which should already contain `connect`.

Let's just name it `config` too as config in this project (except in `<VegaLite/>`)  should mean Voyager's config anyway.

src/components/vega-lite/index.tsx
```diff
@@ -16,6 +17,8 @@ export interface VegaLiteProps {
   logger: Logger;
 
   data: InlineData;
+
+  vegaliteConfig: Config;
```
kanitw: Please just name this `config`.

src/index.tsx
```diff
@@ -4,23 +4,33 @@ import { Provider } from 'react-redux';
...
+import {Config} from 'vega-lite/build/src/config';
 import { Data } from 'vega-lite/build/src/data';
 import {App} from './components/app';
 import { VOYAGER_CONFIG } from './constants';
 import { VoyagerConfig } from './models/config';
 import { configureStore } from './store';
 
 const store = configureStore();
-const config: VoyagerConfig = VOYAGER_CONFIG;
+let config: VoyagerConfig = VOYAGER_CONFIG;
```
kanitw: I think the demo is a bit unnecessary (and should not be here).  More importantly, the default mark color should not be "black".  Please revert this file.

src/reducers/reset.ts
```diff
@@ -13,6 +13,10 @@ export function makeResetReducer<T extends object>(
       // Need to cast as object as TS somehow doesn't know that T extends object already
       const newState = {...state as object} as T;
       Object.keys(resetIndex).forEach((key: keyof T) => {
+        // This is to not reset any property if resetIndex[property] is false.
```
kanitw: Nice catch.  Maybe add a unit test to prevent regression?

src/lib-voyager.tsx
```diff
@@ -220,11 +231,35 @@ export class Voyager {
...
+
+/**
+ * Create an instance of the voyager application.
+ *
+ * @param {Container}       container css selector or HTMLElement that will be the parent
+ *                                    element of the application
+ * @param {Object}          config    configuration options
+ * @param {Array}           data      data object. Can be a string or an array of objects.
+ */
+export function CreateVoyager(container: Container, data: Data, config: VoyagerConfig): Voyager {
```
kanitw: We should not offer two versions of a function with a capital C and a lowercase C.  

So you have two options:

a) Remove `CreateVoyager`.  (Yes, this will break things, but we are currently under pre-release version, so better make things right right now. I'll note in the release note that the next release includes this breaking changes.)  

b) Do not add `createVoyager` with the new parameter since this PR only modifies VoyagerConfig and doesn't really add new parameter to `CreateVoyager` anyway.  

I strongly prefer a) since it's better prepared for the future.  (We also previously let `CreateVoyager` with capital `C` slips in by mistake -- so better fix it before it is too late.)

If you do a), please also help:

1) Update README.md accordingly
2) submit a relevant PR for `voyager-electron`to update relevant API calls

src/lib-voyager.tsx
```diff
@@ -36,7 +45,8 @@ export class Voyager {
   private data: Data;
   private filename: string;
 
-  constructor(container: Container, config: VoyagerConfig, data: Data) {
+  constructor(container: Container, params: VoyagerParams = DEFAULT_VOYAGER_PARAMS) {
```
kanitw: Of the top of my head, I think we don't really need the default `DEFAULT_VOYAGER_PARAMS` as the store already has default `data` and `config`?   (Let me know if I'm wrong as my memory is not perfect.)

Note that if we really need it, we probably better use
```
const {config=DEFAULT_VOYAGER_CONFIG, data=DEFAULT_DATA} = params;
```
so the default is always used if only one parameter is specified.

src/lib-voyager.tsx
ajainarayanan: @kanitw ok makes sense. I added it as a separate function for gradually letting users know that this function will be deprecated in the next release. But since voyager is still in alpha/beta releases I guess it should be ok. Its a simple API. Will revert back.

src/components/plot/index.tsx
ajainarayanan: ok makes sense if we want to extract this as a separate component (or separate repo).

src/lib-voyager.tsx
kanitw: > Will revert back.

Note that if you do b), we will probably still do a) in the future anyway and you'll have to update the API calls.  Maybe it's better if you help us do a) now :p

ajainarayanan: 👍 Will do a). Regarding PR for voyager-electron: There is an already opened PR. Not sure if I can create another one. But will see if I can spare sometime this weekend for it :)

ajainarayanan:
@kanitw Have updated the PR based on review comments. Let me know if I had missed anything.

ajainarayanan:
@kanitw Can you give it a try now. Have added unit tests for reset reducer.

src/lib-voyager.tsx
```diff
@@ -29,10 +29,7 @@ export interface VoyagerParams {
   config?: VoyagerConfig;
 };
 
-const DEFAULT_VOYAGER_PARAMS = {
-  data: {values: ([] as any)},
-  config: DEFAULT_VOYAGER_CONFIG
-};
+const DEFAULT_DATA: any = undefined;
```
kanitw: Remove this?  (We can just not assign default data below.)

src/lib-voyager.test.ui.tsx
```diff
@@ -163,6 +163,48 @@ describe('lib-voyager', () => {
         }
       }, 10);
     });
+    it('CreateVoyager, update config and update data should retain correct config', done => {
```
kanitw: creates a Voyager instance, update ...

src/reducers/reset.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        config: (state: any = defaultState.config) => state,
+        log: (state: any = defaultState.log) => state,
+        shelfPreview: (state: any = defaultState.shelfPreview) => state
+      }),
+      persistentStateToReset,
+      defaultState
+    );
+  };
+
+  it('Should reset to right default value', () => {
```
kanitw: please use "should" (small cap)

src/reducers/reset.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    const resetReducer = getResetReducer();
+    expect(
+      resetReducer(DEFAULT_PERSISTENT_STATE, {type: RESET})
+    ).toEqual(DEFAULT_PERSISTENT_STATE);
+  });
+
+  it('Should reset bookmark when resetIndex is true', () => {
```
kanitw: please use "should" (small cap)

src/reducers/reset.test.ts
```diff
@@ -0,0 +1,83 @@
...
+describe(RESET, () => {
+
+  const persistentStateToReset: ResetIndex<PersistentState> = {
+    bookmark: true,
+    config: false,
+    log: false,
+    shelfPreview: true
+  };
+
+  const getResetReducer = (defaultState = DEFAULT_PERSISTENT_STATE) => {
```
kanitw: Maybe just export persistentReducer in `index.ts` and use it?

ajainarayanan: Since its a unit test for reset reducer just wanted it to be self contained. If we change `persistentReducer` this unit test should be independent of that change as it is testing only reset reducer. May be if I change the model then it can become totally independent.

ajainarayanan:
@kanitw Have updated the PR. Could you give it another try.

ajainarayanan:
@kanitw Any updates on this? Does it require any more changes that I missed?

kanitw:
Sorry I was on a conference and a vacation afterward so I'm a bit slow on this, but I will try to get to this soon.

src/lib-voyager.tsx
```diff
@@ -220,11 +226,11 @@ export class Voyager {
 /**
  * Create an instance of the voyager application.
  *
- * @param {Container} container css selector or HTMLElement that will be the parent
- *                              element of the application
- * @param {Object}    config    configuration options
- * @param {Array}     data      data object. Can be a string or an array of objects.
+ * @param {Container}       container css selector or HTMLElement that will be the parent
+ *                                    element of the application
+ * @param {VoyagerParams}   params    Voyager params. {data, config}. Any new additional parameter
```
kanitw: Voyager parameter object which can contain optional `data` and `config` properties.  

No need to say "Any new additional parameter required by voyager can be added here as named parameter for easier visibilty" as that's not what users of Voyager would do.

src/components/vega-lite/index.tsx
```diff
@@ -72,8 +75,15 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, VegaLiteState>
...
     //   }
     // };
-    const {logger} = this.props;
-    const vlSpec = this.props.spec;
+    const {logger, config} = this.props;
+    let vlSpec = this.props.spec;
+    vlSpec = {
+      ...vlSpec,
+      config: {
+        ...vlSpec.config,
```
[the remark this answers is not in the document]
ajainarayanan: Wouldn't this override the config users pass to modify vega-lite chart's configuration? My understanding was users pass `vegaliteConfig` to override spec configurations for respective charts (for instance modify the color or axis tick format etc.,)

ajainarayanan:
@kanitw Sorry for the delayed response. Was busy at work. Have updated PR based on review comments. Can you give it a try.

ajainarayanan:
Hey @kanitw . Are there any updates on this PR? Did I miss something in my last commit? Can you give another try

src/components/vega-lite/index.tsx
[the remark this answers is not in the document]
kanitw: Yes, it will override since we think about the shared config as a theme.  So suppose you have multiple Vega-Lite charts and you have a common config that you pass in.   Each spec can have config that override the common theme (rather than the other way round). :)

kanitw:
Replied.  Sorry for delay too, I've been quite busy and a bit sick.

src/components/vega-lite/index.tsx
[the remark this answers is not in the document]
ajainarayanan: If that is the case would the common theme ever be applied? Say in the case of bar charts we have defined the fill color to be a blue and if users need to override that would that be possible? I am not sure if I am missing something.

Can you provide an example where a user modifies the fill color of bar charts in voyager with custom config? My assumption was the config user provides are custom config and that would override any default spec config that comes as part of voyager for vega-lite charts.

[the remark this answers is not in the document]
kanitw: > If that is the case would the common theme ever be applied?

Yes, your common theme may have 10 properties and the spec specific theme may have 1 that override the common theme so 9 properties would still be applied. 

> Can you provide an example where a user modifies the fill color of bar charts in voyager with custom config?

Right now there is no such case. But this is how Vega-Lite treat "common" theme too.  So should CompassQL output some config in the future. 

That said, thinking more about it I think I agree with you that if users explicitly specify the config, it should override what's automatically generated by CompassQL.

kanitw:
OK i'm convinced.  The last part needed is to update the Voyager electron app then. :)

ajainarayanan:
@kanitw Thought of updating the voyager-electron app once we have a npm release of voyager. I can try testing with local npm package but would be easier (or complete) if I work against a version of voyager from npm. Let me know if you are thinking of something else. 

##### Note:
 There is an already existing voyager-electron PR for new build changes. Might be easier if I close that and open a new PR with latest voyager changes.

kanitw:
> I can try testing with local npm package

I think it's better to test with local npm package first. :) 

> There is an already existing voyager-electron PR for new build changes. Might be easier if I close that and open a new PR with latest voyager changes.

Do you mean https://github.com/vega/voyager-electron/pull/32?  I guess you can fork from that PR instead of closing it -- otherwise we're losing those changes?

ajainarayanan:
:) I understand. The new PR I was planning to open will have both the changes. Thats why mentioned of closing that PR. Forking a PR from an existing PR becomes a hassle to maintain. Let me see if it makes it simpler.

Issue #731, opened 2017-10-17
More Compact UI for specifying visual encoding


kanitw:
- [ ] Allow empty shelf to have a pop up to select field

- [ ] Design question: if there is already a field.  How do we enable both removing a field and changing a field.

saulshanabrook:
We [started some mockups for more limited Voyager views in Jupyterlab](https://github.com/altair-viz/jupyterlab_voyager/issues/7).

PR #735, opened 2017-10-20
Add Resizable Panes

ssharif6:
Part of #731 

Made data and encoding panes resizable

voyagerresizing

UPDATED GIF:
voyagerresizing

### DIFF
--- a/src/components/app-root/index.tsx
+++ b/src/components/app-root/index.tsx
@@ -35,9 +35,9 @@ class AppRootBase extends React.PureComponent<AppRootProps, {}> {
         bottomPane = <LoadData/>;
       } else {
         bottomPane = (
-          <SplitPane split="vertical" defaultSize={200}>
+          <SplitPane split="vertical" defaultSize={200} minSize={100}>
             <DataPane/>
-            <SplitPane split="vertical" defaultSize={235}>
+            <SplitPane split="vertical" defaultSize={235} minSize={117}>
               <EncodingPane/>
               <ViewPane/>
             </SplitPane>
--- a/src/components/app-root/index.tsx
+++ b/src/components/app-root/index.tsx
@@ -35,9 +35,9 @@ class AppRootBase extends React.PureComponent<AppRootProps, {}> {
         bottomPane = <LoadData/>;
       } else {
         bottomPane = (
-          <SplitPane split="vertical" defaultSize={200} minSize={100}>
+          <SplitPane split="vertical" defaultSize={200} minSize={100} maxSize={200}>
             <DataPane/>
-            <SplitPane split="vertical" defaultSize={235} minSize={117}>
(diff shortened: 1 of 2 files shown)


felixcodes:
LGTM

src/components/app-root/index.tsx
```diff
@@ -35,9 +35,9 @@ class AppRootBase extends React.PureComponent<AppRootProps, {}> {
         bottomPane = <LoadData/>;
       } else {
         bottomPane = (
-          <SplitPane split="vertical" defaultSize={200}>
+          <SplitPane split="vertical" defaultSize={200} minSize={100}>
```
felixcodes: Maybe add `maxSize={200}` as well to prevent users from making the pane wider than the default.

src/components/app-root/index.tsx
```diff
@@ -35,9 +35,9 @@ class AppRootBase extends React.PureComponent<AppRootProps, {}> {
...
             <DataPane/>
-            <SplitPane split="vertical" defaultSize={235}>
+            <SplitPane split="vertical" defaultSize={235} minSize={117}>
```
felixcodes: Maybe add `maxSize={235} as well to prevent users from making the pane wider than the default.

src/components/app-root/index.tsx
kanitw: Not sure why wider that the default is better?

src/components/app-root/index.tsx
kanitw: Not sure why wider that the default is better?

src/components/app-root/index.tsx
```diff
@@ -35,9 +35,9 @@ class AppRootBase extends React.PureComponent<AppRootProps, {}> {
         bottomPane = <LoadData/>;
       } else {
         bottomPane = (
-          <SplitPane split="vertical" defaultSize={200}>
+          <SplitPane split="vertical" defaultSize={200} minSize={100} maxSize={200}>
             <DataPane/>
-            <SplitPane split="vertical" defaultSize={235}>
+            <SplitPane split="vertical" defaultSize={235} minSize={117} maxSize={235}>
```
kanitw: The min size is too small, which allows user to construct a broken state

image

src/components/app.scss
```diff
@@ -132,6 +132,57 @@
     margin: auto;
   }
 
+  // split-pane css
+  .Resizer {
+    background: #000;
```
kanitw: Can you remove this background (as it introduces visual imbalance in our UI)? 

voyager_2_and_vega-lite__git_

kanitw:
Excellent! Thanks for the PR.

felixcodes:
@ssharif6  Congrats on your first PR! 🎉

Issue #634, opened 2017-08-18
Related View Hiding/Tabbing Feature


kanitw:
## Hide (Now)
- [x] separate modification of the store `State` from UI in a single PR

- [x] Try to understand the flow for `dispatchQueries` in `queries/index.ts` and decide if we would dispatch related views request when users hide the related views.

1) If we don't dispatch, then we have to dispatch the request when people enable it.
2) If we still dispatch, then make sure we don't render them until people unhide it again. 

I think 2 makes more sense for now. 

-------

## Tab 

...

kanitw:
@ssharif6 -- feel free to assign this yourself once you accept invitation to join the team.

PR #736, opened 2017-10-26
Add model, action, selector, reducer

ssharif6:
Part of #634

Note: UI code not covered in this PR, just state changes, action, reducer.

* Add logic to toggle hiding related-views
* Added flag in State to represent if related-views are hidden/unhidden
* Add tests for related-view reducer
* Add missing default state and add missing reducer to combineReducers

### DIFF
--- a/src/actions/index.ts
+++ b/src/actions/index.ts
@@ -3,6 +3,7 @@ import {ConfigAction} from './config';
 import {CustomWildcardAction} from './custom-wildcard-field';
 import {DatasetAction} from './dataset';
 import {LogAction} from './log';
+import {RelatedViewAction} from './related-views';
 import {ResetAction} from './reset';
 import {ResultAction} from './result';
 import {ShelfAction} from './shelf';
@@ -34,6 +35,7 @@ export type Action = (
   CustomWildcardAction |
   DatasetAction |
   LogAction |
+  RelatedViewAction |
   ResetAction |
   ResultAction |
   ShelfAction |
@@ -76,6 +78,8 @@ const ACTION_TYPE_INDEX: {[k in ActionType]: 1} = {
   LOG_ERRORS_CLEAR: 1,
   LOG_WARNINGS_ADD: 1,
   LOG_WARNINGS_CLEAR: 1,
+
+  RELATED_VIEW_HIDE_TOGGLE: 1,
 
   RESULT_RECEIVE: 1,
   RESULT_REQUEST: 1,
--- a/src/actions/index.ts
+++ b/src/actions/index.ts
@@ -3,7 +3,7 @@ import {ConfigAction} from './config';
 import {CustomWildcardAction} from './custom-wildcard-field';
 import {DatasetAction} from './dataset';
 import {LogAction} from './log';
-import {RelatedViewAction} from './related-views';
+import {RelatedViewsAction} from './related-views';
 import {ResetAction} from './reset';
 import {ResultAction} from './result';
 import {ShelfAction} from './shelf';
@@ -35,7 +35,7 @@ export type Action = (
   CustomWildcardAction |
   DatasetAction |
   LogAction |
-  RelatedViewAction |
+  RelatedViewsAction |
   ResetAction |
   ResultAction |
   ShelfAction |
@@ -79,7 +79,7 @@ const ACTION_TYPE_INDEX: {[k in ActionType]: 1} = {
   LOG_WARNINGS_ADD: 1,
   LOG_WARNINGS_CLEAR: 1,
 
-  RELATED_VIEW_HIDE_TOGGLE: 1,
+  RELATED_VIEWS_HIDE_TOGGLE: 1,
 
   RESULT_RECEIVE: 1,
   RESULT_REQUEST: 1,
--- a/src/actions/related-views.ts
+++ b/src/actions/related-views.ts
@@ -1,6 +1,6 @@
 import {PlainReduxAction} from './redux-action';
 
-export type RelatedViewAction = HideRelatedViews;
+export type RelatedViewsAction = HideRelatedViews;
 
-export const RELATED_VIEWS_HIDE_TOGGLE = 'RELATED_VIEW_HIDE_TOGGLE';
+export const RELATED_VIEWS_HIDE_TOGGLE = 'RELATED_VIEWS_HIDE_TOGGLE';
 export type HideRelatedViews = PlainReduxAction<typeof RELATED_VIEWS_HIDE_TOGGLE>;
--- a/src/actions/related-views.ts
+++ b/src/actions/related-views.ts
@@ -1,6 +1,6 @@
 import {PlainReduxAction} from './redux-action';
 
-export type RelatedViewAction = HideRelatedView;
+export type RelatedViewAction = HideRelatedViews;
 
-export const RELATED_VIEW_HIDE_TOGGLE = 'RELATED_VIEW_HIDE_TOGGLE';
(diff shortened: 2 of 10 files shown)


src/actions/related-views.ts
```diff
@@ -0,0 +1,6 @@
+import {PlainReduxAction} from './redux-action';
+
+export type RelatedViewAction = HideRelatedView;
+
+export const RELATED_VIEW_HIDE_TOGGLE = 'RELATED_VIEW_HIDE_TOGGLE';
```
kanitw: This is hiding _all_ related view**s** so I wonder if the action should be `RELATED_VIEWS_HIDE_TOGGLE` instead. 

(Similar for `RelatedViewAction` -- however, we may have action that is specific to one kind of related views in the future so I have less strong opinion about the `RelatedViewAction`.)

src/models/index.ts
```diff
@@ -22,6 +23,7 @@ export interface PersistentState {
   bookmark: Bookmark;
   config: VoyagerConfig;
   log: Log;
+  relatedViewToggler: RelatedViewToggler;
```
kanitw: Perhaps just `relatedViews` -- as we may have other related views feature in the future that are not boolean / togglers.

src/models/relatedViewToggler.ts
```diff
@@ -0,0 +1,8 @@
+
+export interface RelatedViewToggler {
+  isRelatedViewHidden: boolean;
```
kanitw: hidden: boolean

src/models/index.ts
```diff
@@ -64,6 +66,7 @@ export const DEFAULT_PERSISTENT_STATE: PersistentState = {
   bookmark: DEFAULT_BOOKMARK,
   config: DEFAULT_VOYAGER_CONFIG,
   log: DEFAULT_LOG,
+  relatedViews: DEFAULT_RELATED_VIEW_TOGGLER,
```
kanitw: DEFAULT_RELATED_VIEWS

src/models/relatedViewToggler.ts
```diff
@@ -0,0 +1,8 @@
+
+export interface RelatedViewToggler {
```
kanitw: `RelatedViews` 

Also rename the file and all other relevant things

src/reducers/index.test.ts
```diff
@@ -76,6 +77,7 @@ describe('reducers/index', () => {
       expect(selectBookmark(state)).toEqual(DEFAULT_BOOKMARK);
       expect(selectCustomWildcardFields(state)).toEqual(DEFAULT_CUSTOM_WILDCARD_FIELDS);
       expect(selectDataset(state)).toEqual(DEFAULT_DATASET);
+      expect(selectRelatedViewToggler(state)).toEqual(DEFAULT_RELATED_VIEW_TOGGLER);
```
kanitw: selectRelatedViews

src/reducers/index.ts
```diff
@@ -86,21 +88,22 @@ const persistentStateToReset: ResetIndex<PersistentState> = {
...
   log: false,
+  relatedViews: true,
   shelfPreview: true
 };
 
 const persistentReducer = makeResetReducer(
   combineReducers<PersistentState>({
     bookmark: bookmarkReducer,
     config: configReducer,
+    relatedViewToggler: relatedViewReducer,
```
kanitw: relatedViews

src/reducers/related-view.test.ts
```diff
@@ -0,0 +1,35 @@
...
+        isHidden: false
+      },
+      {
+        type: RELATED_VIEWS_HIDE_TOGGLE
+      }
+    )).toEqual(expectedRelatedViews);
+  });
+});
+
+describe('reducers/related-views', () => {
```
kanitw: Usually, we have just one 'describe' for multiple test cases that are under the same hierarchy

src/reducers/related-views.ts
```diff
@@ -0,0 +1,10 @@
+import {Action} from '../actions/index';
+import {DEFAULT_RELATED_VIEWS, RelatedViews} from '../models/related-views';
+
+export function relatedViewsReducer(relatedViewToggler: RelatedViews = DEFAULT_RELATED_VIEWS,
+                                    action: Action): RelatedViews {
+  const {isHidden} = relatedViewToggler;
```
kanitw: Don't you have to check for action?  Wouldn't this mean that all actions will toggle the related view?

src/reducers/related-views.ts
```diff
@@ -0,0 +1,10 @@
...
+  return {
+    isHidden: !isHidden
```
kanitw: 2) Why do we return `isHidden: !isHidden`, not just `isHidden`?

src/actions/related-views.ts
```diff
@@ -0,0 +1,6 @@
+import {PlainReduxAction} from './redux-action';
+
+export type RelatedViewAction = HideRelatedViews;
```
kanitw: RelatedViewsAction

src/actions/index.ts
```diff
@@ -77,6 +79,8 @@ const ACTION_TYPE_INDEX: {[k in ActionType]: 1} = {
   LOG_WARNINGS_ADD: 1,
   LOG_WARNINGS_CLEAR: 1,
 
+  RELATED_VIEW_HIDE_TOGGLE: 1,
```
kanitw: RELATED_VIEWS_HIDE_TOGGLE

src/reducers/related-views.ts
kanitw: I also wonder if just `hidden` is a better name than `isHidden`, but I don't have strong preference about this part.

felixcodes: @kanitw Please correct me if I'm misunderstanding -- but I believe the `relatedViewToggler` parameter contains the current state before the action is pressed.  The value of `isHidden` is derived from `relatedViewToggler.isHidden` -- i.e. the current value of isHidden. `RelatedViewsAction` should flip the value of `isHidden` and returning `{isHidden}` as per your suggestion wouldn't flip the value at all.

@ssharif6 Is this what you were thinking?

ssharif6: @felixcodes Yes. I was thinking along the lines of toggling the `isHidden` value. So a `isHidden: !isHidden` would flip from displaying hidden views in the state, and then not displaying them.

kanitw: That make sense. Still wouldn't you need an action?

felixcodes: Yes, @ssharif6 is missing a 
switch (action.type) case RelatedViewsAction:

felixcodes: case RELATED_VIEWS_HIDE_TOGGLE:*** (on mobile, can't edit comments)

ssharif6: Yes! That is definitely true and is being added for next iteration

src/reducers/related-views.ts
```diff
@@ -0,0 +1,17 @@
+import { Action } from '../actions/index';
+import { RELATED_VIEWS_HIDE_TOGGLE } from '../actions/related-views';
+import { DEFAULT_RELATED_VIEWS, RelatedViews } from '../models/related-views';
+
+export function relatedViewsReducer(relatedViewToggler: RelatedViews = DEFAULT_RELATED_VIEWS,
+                                    action: Action): RelatedViews {
```
kanitw: FYI, I actually don't like when a code is tabbed by half a page.  Perhaps better to tab by two or do

```
export function relatedViewsReducer(
  relatedViewToggler: RelatedViews = DEFAULT_RELATED_VIEWS,
  action: Action
): RelatedViews {
```

PR #738, opened 2017-10-29
Fix missing filename bug

### LINKED ISSUE #715: Filename is missing
Filename is missing for default datasets, uploaded local CSV data, and remote data loaded via URL.

image

felixcodes:
Fix #715 
Filenames work again :)

And thanks to @ssharif6 for digging around with me last night.

image

### DIFF
--- a/src/actions/dataset.ts
+++ b/src/actions/dataset.ts
@@ -60,17 +60,19 @@ export function datasetLoad(name: string, data: Data): DatasetLoad {
         .then(response => response.json())
         .catch(errorCatch)
         .then((values: any) => {
-          return buildSchemaAndDispatchDataReceive({values}, config, dispatch);
+          return buildSchemaAndDispatchDataReceive({values}, config, dispatch, name);
         });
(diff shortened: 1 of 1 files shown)


src/actions/dataset.ts
```diff
@@ -60,17 +60,19 @@ export function datasetLoad(name: string, data: Data): DatasetLoad {
...
+      return buildSchemaAndDispatchDataReceive(data, config, dispatch, name);
     } else {
       throw new Error('dataset load error: dataset type not detected');
     }
   };
 };
 
-function buildSchemaAndDispatchDataReceive(data: InlineData, config: VoyagerConfig, dispatch: Dispatch<Action>) {
+function buildSchemaAndDispatchDataReceive(
+  data: InlineData, config: VoyagerConfig, dispatch: Dispatch<Action>, name: String
```
kanitw: Please use `string` rather than String :) 

https://stackoverflow.com/questions/14727044/typescript-difference-between-string-and-string

PR #756, opened 2018-01-27
Add getSpec handler to lib-voyager

### LINKED ISSUE #746: Add getSpec function to Voyager class
I'd like to add an API like this to https://github.com/davidanthoff/DataVoyager.jl, the code below is how a user would start and interact with voyager from the julia REPL:
````julia
julia> using DataVoyager, RDatasets

julia> v = dataset("ggplot2", "mpg") |> Voyager()
````
This opens up the voyager UI in an electron window and you can work with it. Now fiddle with the Voyager UI, when you are done, enter the following command on the REPL to save the spec:
````julia
julia> v |> save("myspec.vegalite")
````
All I would need for that is some way to extract the vega-lite spec of the specified view from the ``Voyager`` instance. An added bonus would be if the function would accept a flag whether I want the data to be included or not in the spec.

### PR DESCRIPTION
**Pull request by ssharif6**
Add `getSpec` to lib-voyager to export Vega-Lite spec for current specified view.

`getSpec` accepts a boolean value indicating whether or not to attach data payload (inline) in request. 

Fix #746

### DIFF
--- a/src/lib-voyager.test.ui.tsx
+++ b/src/lib-voyager.test.ui.tsx
@@ -225,7 +225,7 @@ describe('lib-voyager', () => {
       }, DEFAULT_TIMEOUT_LENGTH);
     });
 
-    it('getSpec returns Vega-Lite spec', done => {
+    it('getSpec returns Vega-Lite spec without inline data', done => {
       setTimeout(() => {
         try {
           const voyagerInst = CreateVoyager(container, undefined, undefined);
@@ -257,7 +257,7 @@ describe('lib-voyager', () => {
 
           setTimeout(() => {
             try {
-              const retrievedSpec = voyagerInst.getSpec();
+              const retrievedSpec = voyagerInst.getSpec(false);
               expect(retrievedSpec).toEqual({
                 data: {
                   name: 'source'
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by davidanthoff** [comment]
Great, much appreciated!

File: src/lib-voyager.test.ui.tsx [hunk 1]
```diff
@@ -224,5 +224,80 @@ describe('lib-voyager', () => {
...
+            }
+          };
+          voyagerInst.setSpec(spec);
+
+          setTimeout(() => {
+            try {
+              const retrievedSpec = voyagerInst.getSpec();
+              expect(retrievedSpec).toEqual({
+                data: {
+                  name: 'source'
```
felixcodes: cc @kanitw: Every VL spec in the store has `data: {name: 'source'}`. Should other programs like David's receive the spec like this when they make a getSpec() API call?

kanitw: Great question, I think appending data (either url or inline) might make more sense?

ssharif6: Good suggestion! I recall @davidanthoff mentioning the idea of having a boolean flag indicating whether or not to attach the data payload. I made this change in the next iteration.

src/lib-voyager.test.ui.tsx
```diff
@@ -224,5 +224,157 @@ describe('lib-voyager', () => {
 
       }, DEFAULT_TIMEOUT_LENGTH);
     });
+
+    it('getSpec returns Vega-Lite spec without inline data', done => {
```
kanitw: getSpec with ..(parameter name).. = false returns ...

src/lib-voyager.test.ui.tsx
```diff
@@ -224,5 +224,157 @@ describe('lib-voyager', () => {
...
+
+        } catch (err) {
+          done.fail(err);
+        }
+
+      }, DEFAULT_TIMEOUT_LENGTH);
+    });
+  });
+
+  it('getSpec returns Vega-Lite spec with inline data', done => {
```
kanitw: getSpec with ..(parameter name).. = false returns ...

src/lib-voyager.tsx
```diff
@@ -147,6 +149,23 @@ export class Voyager {
...
+   * Gets Vega-Lite spec of current specified view
+   *
+   * @returns {Readonly<Spec>}
+   *
+   * @memberof Voyager
+   */
+  public getSpec(includeData: boolean): FacetedCompositeUnitSpec {
+    const spec = selectMainSpec(this.store.getState());
+    if (includeData) {
+      spec.data = selectData(this.store.getState());
```
kanitw: Would this cause side effect to the object in the store?  I'd say better 

````
return {
  data:selectData(this.store.getState(), 
  ...spec
};
```

src/lib-voyager.test.ui.tsx
davidanthoff: Yes, a flag is ideal there. I'm not sure, but the current version now would include the ``data`` element with the name of the source if the function is called with the flag to not include data? Or did I misunderstand that? I guess what I'm saying is that when the function is called with the flag to not include data, it would be great if the spec just didn't have any `data` element.

kanitw: > the current version now would include the data element with the name of the source if the function is called with the flag to not include data? 

Yes, that's the right behavior.  A spec without `data` is an invalid Vega-Lite spec and would make it fail to match output type.  This way you can directly bind data to the named data source. If you want to replace `data` with your own new data definition, as a user of the API, you can always replace `data` with what you want later.

davidanthoff: Ah, got it, that makes sense!

davidanthoff:
Cool, thanks so much everyone! Is there a chance that you could release a new alpha? Then I could start to use this in the julia wrapper for this.

felixcodes:
@davidanthoff alpha.14 is released!

davidanthoff:
Perfect, thanks!

PR #765, opened 2018-02-04
Remove 'protected' for life cycle methods in VegaLite module

### LINKED ISSUE #694: Protected lifecycle methods not compatible with @types/react >= 15.0.25
Voyager is using `protected` lifecycle methods ([example](https://github.com/vega/voyager/blob/master/src/components/vega-lite/index.tsx#L68)), but starting from `@types/react v15.0.25`  all of the lifecycle methods are forced to be `public`.

This problem emerged when I was trying to use the `Vega-Lite` component in my own project (see [stack overflow question](https://stackoverflow.com/a/46038869/1270459)). I fixed my problem by downgrading `@types/react` version, but I wonder if voyager will make the lifecycle methods `public` in the future.

Relevant issue: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/16893

ZeningQu:
Fix #694

### DIFF
--- a/src/components/vega-lite/index.tsx
+++ b/src/components/vega-lite/index.tsx
@@ -49,6 +49,68 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, VegaLiteState> 
     );
   }
 
+  public componentDidMount() {
+    if (this.mountTimeout) {
+      clearTimeout(this.mountTimeout);
+    }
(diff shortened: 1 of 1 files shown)


kanitw:
Thanks @felixcodes  ;)

kanitw:
and @ZeningQu  :)

PR #768, opened 2018-02-13
Fix a test in lib-voyager.test.ui.tsx by expecting data `_id` added by Vega

ZeningQu:
It seems that `vega` is adding a field called `_id` to data values (https://vega.github.io/vega/docs/data/), and that caused a test case to fail. I'm fixing it by adding `expect.any(Number)` in the expected `retrievedSpec`


ZeningQu:
For some reason Travis CI has inconsistent test result than my local `yarn test` results. So I'm closing this PR now.

Issue #686, opened 2017-09-01
`src/reducers/index.test.ts` fail to run on Travis


kanitw:
![image](https://user-images.githubusercontent.com/111269/29989967-983e66ce-8f2b-11e7-8d91-0c29f2c7838a.png)
```
  ● Test suite failed to run
    ProcessTerminatedError: cancel after 2 retries!
      
      at Farm.<anonymous> (node_modules/worker-farm/lib/farm.js:81:25)
      at Array.forEach (native)
      at Farm.<anonymous> (node_modules/worker-farm/lib/farm.js:75:36)
      at ontimeout (timers.js:386:14)
      at tryOnTimeout (timers.js:250:5)
      at Timer.listOnTimeout (timers.js:214:5)
A worker process has quit unexpectedly! Most likely this is an initialization error.
pec in JSON and, if applicable, screenshots or GIF videos (e.g., using https://www.cockos.com/licecap/)
```

ZeningQu:
Had the same issue today with `actions/index.test.ts`

![screen shot 2018-02-13 at 3 11 11 pm](https://user-images.githubusercontent.com/822034/36179190-2fe2c346-10d0-11e8-95d8-7ff8fd822412.png)

```
FAIL  src/actions/index.test.ts
  ● Test suite failed to run
    ProcessTerminatedError: cancel after 2 retries!
      
      at Farm.<anonymous> (node_modules/worker-farm/lib/farm.js:81:25)
      at Array.forEach (native)
      at Farm.<anonymous> (node_modules/worker-farm/lib/farm.js:75:36)
      at ontimeout (timers.js:386:11)
      at tryOnTimeout (timers.js:250:5)
      at Timer.listOnTimeout (timers.js:214:5)
A worker process has quit unexpectedly! Most likely this is an initialization error.
npm ERR! Linux 4.14.12-041412-generic
npm ERR! argv "/home/travis/.nvm/versions/node/v6.13.0/bin/node" "/home/travis/.nvm/versions/node/v6.13.0/bin/npm" "run" "codecov"
npm ERR! node v6.13.0
npm ERR! npm  v3.10.10
npm ERR! code ELIFECYCLE
npm ERR! datavoyager@2.0.0-alpha.15 codecov: `jest --no-cache --coverage && codecov`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the datavoyager@2.0.0-alpha.15 codecov script 'jest --no-cache --coverage && codecov'.
[...]

PR #769, opened 2018-02-13
Make view and runView() protected in VegaLite class 

ZeningQu:
because I want to extend this `VegaLite` class in my own app so that whenever `runView()` is called, the vega view will be updated in my app state and I can run further consistency analysis on the rendered view.

- [x] Change `view` from `private` to `protected`
- [x] Change `runView()` from `private` to `protected`
- [x] Move `runView()` up because protected methods should come before private methods

### DIFF
--- a/src/components/vega-lite/index.tsx
+++ b/src/components/vega-lite/index.tsx
@@ -25,7 +25,7 @@ export interface VegaLiteState {
 const CHART_REF = 'chart';
 
 export class VegaLite extends React.PureComponent<VegaLiteProps, VegaLiteState> {
-  private view: vega.View;
+  protected view: vega.View;
   private size: {width: number, height: number};
 
   private mountTimeout: number;
@@ -151,6 +151,13 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, VegaLiteState> 
     }
   }
 
+  protected runView() {
+    try {
+      this.view.run();
+    } catch (err) {
+      this.props.logger.error(err);
+    }
+  }
 
   private bindData() {
     const {data, spec} = this.props;
@@ -160,14 +167,6 @@ export class VegaLite extends React.PureComponent<VegaLiteProps, VegaLiteState> 
             .remove(() => true) // remove previous data
             .insert(data.values)
       );
-    }
-  }
-
-  private runView() {
-    try {
-      this.view.run();
-    } catch (err) {
-      this.props.logger.error(err);
     }
   }
 


kanitw:
I doubt if this is the right move.    
The VegaLite class should update automatically whenever the Vega-Lite input spec change.  

In the case that your spec doesn't change, even if you call runView() it won't change any result?

ZeningQu:
Yeah it won't change the view, but I want to write the `view` into my app state, and in my own app I want to overwrite `runView()` so that whenever the view is updated, my app state will update as well. Kind of like below:

```
import { VegaLite, VegaLiteProps } from 'datavoyager/build/components/vega-lite';
import * as vega from 'vega';

export class VegaLiteWrapper extends VegaLite {
  constructor(props: VegaLiteProps) {
    super(props);
  }

  protected runView() {
    try {
      this.view.run();
      this.view.runAfter((renderedView: vega.View) => {
        // write renderedView._runtime.scales into store state, for consistency analysis
      });
    } catch (err) {
      this.props.logger.error(err);
    }
  }
}
```

ZeningQu:
Test is failing because of this issue: https://github.com/vega/voyager/issues/686

ZeningQu:
> I see. This seems a bit weird as view doesn't suppose to cause change to the state, but I can't think of other way right now anyway.

Yeah, I understand that normally view should not change state. View should read state. And user interactions should change state. 

But I thought if (in)consistency detection is all about detecting inconsistencies in the rendered views, then views should be written and updated in the state.

I thought about doing (in)consistency detection on the specs only, or re-constructing the views from the specs, but they all seem less accurate than getting the actual rendered view.

kanitw:
> I thought about doing (in)consistency detection on the specs only, or re-constructing the views from the specs, but they all seem less accurate than getting the actual rendered view.

Not sure why is that less accurate?   For example, you could separately run `vl.compile` and still get the Vega output of a VL spec, and analyze the spec specifications of the Vega output.

ZeningQu:
Yeah but Vega spec doesn't have the computed scale `domain` and `range` which I need for my consistency analysis. The scale `domain` depends on the spec, the raw data domain, and the filters created by user in Voyager. If I reconstruct the view I need to combine all these sources, so I prefer to get the actual scale domain from the actual view that's rendered.

kanitw:
Okie.   computed domain makes sense.  @felixcodes Could you help @ZeningQu with why test is failing?

ZeningQu:
I just found out that React doesn't like me to inherit the `VegaLite` component, but suggests I should pass in `props` to `VegaLite` for anything I want to override [1, 2]. 

So I think I will:
- undo the `private` -> `protected` changes I made
- add an optional prop `viewWriter: () => void` prop instead.

[1] https://github.com/facebook/react/issues/8774

[2] https://reactjs.org/docs/composition-vs-inheritance.html#specialization

kanitw:
Cool.  I'll close this issue then.

ZeningQu:
Thank you both @kanitw and @felixcodes 
I will open a new PR when it's ready!

PR #776, opened 2018-02-26
Add viewRunAfterCallback in VegaLite props

ZeningQu:
- [x] Add an optional prop called `viewRunAfterCallback` to the VegaLite component so that I can get the vega view from this component after it has been run.

### DIFF
--- a/src/components/vega-lite/index.tsx
+++ b/src/components/vega-lite/index.tsx
@@ -16,6 +16,8 @@ export interface VegaLiteProps {
   logger: Logger;
 
   data: InlineData;
+
+  viewRunAfter?: (view: vega.View) => any;
 }
 
 export interface VegaLiteState {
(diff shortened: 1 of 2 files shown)


src/components/vega-lite/index.tsx
```diff
@@ -16,6 +16,8 @@ export interface VegaLiteProps {
   logger: Logger;
 
   data: InlineData;
+
+  viewListener?: (view: vega.View) => any;
```
kanitw: Are there other possible even that can be listen in Vega? 

If so, this could be viewRunAfter / viewRunAfterListener or something just in case we add other view listeners in the future?

ZeningQu:
@kanitw I renamed the property `viewRunAfterCallback` -- it's a bit verbose, but it's consistent with the [vega view run api](https://github.com/vega/vega-view#view_run): this is the callback that gets called in `vega.view.runAfter()`

kanitw:
perhaps just `viewRunAfter`?

PR #778, opened 2018-02-27
Stop query compassQL when related-views tab hidden

ssharif6:
Stop compassQL for querying related views when related-views toggle mode is on.

voyagerhidetoggle


ssharif6:
@kanitw Done!

Issue #782, opened 2018-03-30
Debounce for PropertyEditor


ssharif6:
This issue is closely related to #757 (Custom Encoding)

Currently, in the property-editor component, the `changeFieldProperty` action is called when the React-Form's `onChange` method is invoked. 

This works ok for select tags, but for text input and sliders, it will render whenever there is a change, which is problematic (If i try to change an axis label, after you type one character, it will make the change with just that first character, etc)

One method is to implement a debounce method so that there is a small delay after the user has stopped typing/interacting with the encoding types.

Another possible solution is instead of using `onChange`, we call our Redux Action via the `onSubmit` handler.

Issue #757, opened 2018-01-29
Visualization Customization


felixcodes:

curran:
@felixcodes What does this issue mean?

kanitw:
TODO:

Mock 

- Position (x,y):  scale, axis, stack, sort, ...

- Facet:  header, sort? (maybe not very complete yet)

- Mark Property (color, size, shape): scale, legend, sort, ...

- Detail -- highly suspicious that nothing is needed 

- Text: just format

kanitw:
https://github.com/vega/voyager/pull/789 is a good start, but we also have to: 

- [ ] Add all properties for position
- [ ] Add all properties for mark property channel 
- [ ] Add all properties for text 
- [ ] Add all properties for facet header 

## Advanced Rules 

Here are a few that I notice: 

- [ ] stack shouldn't always appear
- [ ] scale type shouldn't always list all scale types (list only valid types)

PR #789, opened 2018-04-05
Custom Encoding for Positional Channels

ssharif6:
Fix: #782 

This issue is the skeleton for the custom encoding issues (#757)

This PR allows users to customize encoding for positional fields from the property-editor in a tabbed-view, where each tab represents an encoding type.  

Gifs:

Quantitative Fields:
fieldcustomizer

Non-Quantitative Fields:
customencodingnonquant


src/models/property-editor.ts
```diff
@@ -0,0 +1,137 @@
+import * as vlSchema from 'vega-lite/build/vega-lite-schema.json';
```
ssharif6: Where should the custom schemas be located? Wasn't sure where to put the custom schemas so put into `/models` for now

src/components/encoding-pane/field-customizer.tsx
```diff
@@ -12,35 +12,90 @@ export interface FieldCustomizerProps extends ActionHandler<SpecEncodingAction>
   fieldDef: ShelfFieldDef;
 }
 
+export interface CustomProp {
+  prop: string;
+  nestedProp: string;
+}
 
 export class FieldCustomizerBase extends React.PureComponent<FieldCustomizerProps, {}> {
+  private static channelEncodingMap() {
```
ssharif6: Not too sure where to put this encodingMap, but is easily refactorable

src/models/property-editor.ts
```diff
@@ -0,0 +1,137 @@
...
+    "ui:placeholder": "auto",
+    "ui:emptyValue": "auto"
+  }
+};
+
+export const SCHEMA_OBJ = {
+  "scale_type_quantitative": SCALE_TYPE_SCHEMA,
+  "axis_orient_quantitative": AXIS_ORIENT_SCHEMA,
+  "axis_title_quantitative": AXIS_TITLE_SCHEMA,
+  "color_fill_quantitative": COLOR_COLOR_SCHEMA,
```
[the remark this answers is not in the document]
kanitw: (I don't think the schema for value should be mixed up with fieldDef's properties.)

src/models/property-editor.ts
```diff
@@ -0,0 +1,137 @@
+import * as vlSchema from 'vega-lite/build/vega-lite-schema.json';
+
+export const COLOR_COLOR_SCHEMA = {
+  "type": "object",
+  "properties": {
+    "colorPicker": {
```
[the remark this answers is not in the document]
ssharif6: Not yet. This will be needed for mark channels, which isn't part of this PR, so i'll delete it

src/components/encoding-pane/property-editor.tsx
```diff
@@ -14,26 +15,32 @@ export interface PropertyEditorProps extends ActionHandler<SpecEncodingAction> {
   fieldDef: ShelfFieldDef;
 }
 
-export class PropertyEditorBase extends React.PureComponent<PropertyEditorProps, {}> {
+export interface PropertyEditorState {
+  formData: any;
```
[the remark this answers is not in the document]
ssharif6: The React-Form that we're using stores the formData result as arbitrary JSON, so there's no easy way to be type-safe in this situation, especially since the schema of the formData result can be very complex/convoluted.

src/components/encoding-pane/property-editor.tsx
```diff
@@ -14,26 +15,32 @@ export interface PropertyEditorProps extends ActionHandler<SpecEncodingAction> {
   fieldDef: ShelfFieldDef;
 }
 
-export class PropertyEditorBase extends React.PureComponent<PropertyEditorProps, {}> {
+export interface PropertyEditorState {
```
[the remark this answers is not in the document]
kanitw: How do you make sure that the component's internal state is in sync with the spec in the store?
(Imagine a case where you load a spec into Voyager and it already has log scale.) 

Instead of keeping internal state, better read directly from the store and derive a sensible `formData` from the stored value?  

That said, you know more about this infrastructure, so I might have overlooked something here.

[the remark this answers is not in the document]
ssharif6: This comment makes sense. I'll try to address this in the next pass!

src/models/property-editor.ts
```diff
@@ -0,0 +1,137 @@
...
+  },
+};
+
+export const SCALE_TYPE_SCHEMA = {
+  "type": "object",
+  "title": "Scale",
+  "properties": {
+    "scaleTypeSelect": {
+      "title": "Type",
+      "enum": (vlSchema as any).definitions.ScaleType.enum,
```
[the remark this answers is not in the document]
ssharif6: vlSchema is currently a massive JSON blob, so i'm not sure how we'd get around this

src/models/property-editor.ts
```diff
@@ -0,0 +1,137 @@
...
+      "maximum": 100
+    },
+  },
+};
+
+export const SCALE_TYPE_SCHEMA = {
+  "type": "object",
+  "title": "Scale",
+  "properties": {
+    "scaleTypeSelect": {
```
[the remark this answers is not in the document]
ssharif6: The name is used to identity/connect the schema and UISchema together. So `scaleTypeSelect` is referenced in the UiSchema `SCALE_TYPE_UISCHEMA`

src/components/encoding-pane/field-customizer.tsx
```diff
@@ -12,35 +14,136 @@ export interface FieldCustomizerProps extends ActionHandler<SpecEncodingAction>
...
+      nestedProp: 'orient'
+    },
+    {
+      prop: 'axis',
+      nestedProp: 'title'
+    }
+  ]
+};
+
+// TODO: Mark channel encoding
```
kanitw: Why TODO later?

src/components/encoding-pane/field-customizer.tsx
```diff
@@ -12,35 +14,136 @@ export interface FieldCustomizerProps extends ActionHandler<SpecEncodingAction>
...
+const CHANNEL_ENCODING_MAP = {
```
kanitw: This name is also NOT descriptive.  

At suggested earlier, having a map/index here will restrict the type of properties that affect the field property index.  Why don't you just have switch-case or if-else inside the `getEncodingTypeForChannel` method (the method should be renamed though)?

src/components/encoding-pane/field-customizer.tsx
```diff
@@ -12,35 +14,136 @@ export interface FieldCustomizerProps extends ActionHandler<SpecEncodingAction>
...
+                return (
+                  <Tab key={i}>{encodingType}</Tab>
+                );
+              })
+            }
+          </TabList>
+          <div>
+            {
+              keys.map((encodingType, i) => {
+                const customProps = this.getEncodingTypeForChannel()[encodingType] as CustomProp[];
```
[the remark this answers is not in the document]
kanitw: Also, why calling `this.getEncodingTypeForChannel()` twice in the same method. This seems redundant.

src/components/encoding-pane/field-customizer.tsx
```diff
@@ -12,35 +14,136 @@ export interface FieldCustomizerProps extends ActionHandler<SpecEncodingAction>
...
     );
   }
 
-  private customizableProps() {
-    return [{
-      prop: 'scale',
-      nestedProp: 'type'
-    }];
+  // TODO: Refactor else condition from just mark
+  private getEncodingTypeForChannel() {
```
kanitw: I found the term "encoding type" is very confusing. 

Type = Quantitative / Nominal / ...

Encoding = mapping between encoding channels (x/y/color/...) and fields. 

Overloading another terms for "type" and "encoding" is probably not a good idea. 

Also why "for channel"?  The return result depends on channel, but also depend on other properties (e.g., `type`, for now). 

These are property groups/categories.  So name it `getFieldPropertyGroupIndex`?

src/components/encoding-pane/property-editor.tsx
```diff
@@ -43,9 +62,22 @@ export class PropertyEditorBase extends React.PureComponent<PropertyEditorProps,
     );
   }
 
+  // TODO: Stack should eventually be part of FieldDef
+  protected constructFormData() {
```
kanitw: `private get formdata()` ?

src/components/encoding-pane/property-editorSchema.ts
```diff
@@ -0,0 +1,112 @@
+import * as vlSchema from 'vega-lite/build/vega-lite-schema.json';
+
+const STACK_STACK_SCHEMA = {
```
kanitw: Why STACK_STACK?  Stack is there is no `stack.stack`

src/components/encoding-pane/property-editorSchema.ts
```diff
@@ -0,0 +1,112 @@
...
+
+const STACK_STACK_UISCHEMA = {
+  "stackSelect": {
+    "ui:widget": "select",
+    "ui:placeholder": "auto",
+    "ui:emptyValue": "auto"
+  }
+};
+
+export const SCHEMA_OBJ = {
```
kanitw: Can you get rid of all mapping schema like this as it won't scale well?

src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -72,7 +72,8 @@ class EncodingShelfBase extends React.PureComponent<
             attachment="top left"
             targetAttachment="bottom left"
           >
-            {(fieldDef && !isWildcardChannelId(id)) ?
+            {(fieldDef && !isWildcardChannelId(id) && (this.props.id.channel === 'x' ||
```
ssharif6: Disable the drop-down for channels that are not positional, since property-editor is not supported for those channels yet.

src/components/encoding-pane/field-customizer.tsx
```diff
@@ -12,35 +14,141 @@ export interface FieldCustomizerProps extends ActionHandler<SpecEncodingAction>
...
+      nestedProp: 'orient'
+    },
+    {
+      prop: 'axis',
+      nestedProp: 'title'
+    }
+  ]
+};
+
+const POSITION_FIELD_TEMPORAL_INDEX = {
```
ssharif6: Refactor to separate index. Looks identical to `POSITION_FIELD_NOMINAL_INDEX` currently, but in the future can have different properties

src/components/encoding-pane/encoding-shelf.tsx
kanitw: Why commenting in the PR?  If you have to comment in a PR, that's a sign that you better add comment to the code? 

(Commenting in the PR is better for asking me questions.)

kanitw: you can refer to `id` without `this.props` too.

src/components/encoding-pane/field-customizer.tsx
```diff
@@ -12,35 +14,141 @@ export interface FieldCustomizerProps extends ActionHandler<SpecEncodingAction>
...
+      prop: 'scale',
+      nestedProp: 'type'
+    },
+    {
+      prop: 'axis',
+      nestedProp: 'title'
+    },
+    {
+      prop: 'stack',
+      nestedProp: undefined
```
kanitw: make `nestedProp` optional and drop this line?

src/components/encoding-pane/field-customizer.tsx
```diff
@@ -12,35 +14,141 @@ export interface FieldCustomizerProps extends ActionHandler<SpecEncodingAction>
...
-      prop: 'scale',
-      nestedProp: 'type'
-    }];
+  private getFieldPropertyGroupIndex() {
+    const {shelfId, fieldDef} = this.props;
+    if (shelfId.channel === Channel.X || shelfId.channel === Channel.Y) {
+      switch (fieldDef.type) {
+        case ExpandedType.QUANTITATIVE:
+          return POSITION_FIELD_QUANTITATIVE_INDEX;
+        case ExpandedType.NOMINAL || ExpandedType.ORDINAL:
```
kanitw: I don't think `||` works in `switch`-`case`. 

```
case ExpandedType.NOMINAL:
case ExpandedType.ORDINAL:
```

src/components/encoding-pane/property-editor.tsx
```diff
@@ -1,11 +1,16 @@
...
+import {debounce} from 'throttle-debounce';
+import {ActionHandler} from '../../actions';
+import {SPEC_FIELD_NESTED_PROP_CHANGE, SPEC_FIELD_PROP_CHANGE, SpecEncodingAction} from '../../actions/shelf';
+import {ShelfFieldDef, ShelfId} from '../../models/shelf/spec';
 import * as styles from './property-editor.scss';
+import {
+  AXIS_ORIENT_SCHEMA, AXIS_ORIENT_UISCHEMA,
+  AXIS_TITLE_SCHEMA, AXIS_TITLE_UISCHEMA,
+  SCALE_TYPE_SCHEMA, SCALE_TYPE_UISCHEMA, STACK_SCHEMA, STACK_UISCHEMA,
+} from './property-editorSchema';
```
kanitw: filename should have lower cases as Mac is natively case insensitive   => `property-editor-schema`.

src/components/encoding-pane/field-customizer.tsx
kanitw: If it looks identical right now, don't separate as it's hard to read. Plus, if they differ in the future, they will share common parts that should be declared once.

src/components/encoding-pane/property-editorSchema.ts
```diff
@@ -0,0 +1,79 @@
...
+      "type": "string"
+    }
+  }
+};
+
+export const SCALE_TYPE_SCHEMA = {
+  "type": "object",
+  "title": "Scale",
+  "properties": {
+    "scaleTypeSelect": {
```
kanitw: why not "type"?

src/components/encoding-pane/property-editorSchema.ts
```diff
@@ -0,0 +1,79 @@
...
+      "title": "Title"
+    },
+  },
+};
+
+export const STACK_SCHEMA = {
+  "type": "object",
+  "title": "Stack",
+  "properties": {
+    "stackSelect": {
```
kanitw: why not `stack`?

src/components/encoding-pane/property-editorSchema.ts
```diff
@@ -0,0 +1,79 @@
+import * as vlSchema from 'vega-lite/build/vega-lite-schema.json';
+
+export const AXIS_TITLE_SCHEMA = {
+  "type": "object",
+  "title": "Axis",
+  "properties": {
+    "axisTitle": {
```
kanitw: What's the naming convention for the properties of schema here? 

They seem inconsistent.  axisTitle, stackSelect, ... 

Define a consistent policy.  Add comment what's your policy at the top of the file.

src/components/encoding-pane/property-editorSchema.ts
```diff
@@ -0,0 +1,79 @@
+import * as vlSchema from 'vega-lite/build/vega-lite-schema.json';
+
+export const AXIS_TITLE_SCHEMA = {
```
kanitw: Is this AXIS_TITLE_SCHEMA or AXIS_SCHEMA?

src/components/encoding-pane/property-editorSchema.ts
```diff
@@ -0,0 +1,79 @@
...
+      "type": "string"
+    }
+  }
+};
+
+export const AXIS_ORIENT_SCHEMA = {
+  "type": "object",
+  "title": "Axis",
+  "properties": {
+    "orient": {
```
kanitw: Why isn't this combined in the same schema as `axisTitle`?

Why is it called "orient", not "axisOrient"?

src/components/encoding-pane/property-editorSchema.ts
```diff
@@ -0,0 +1,79 @@
...
+  "properties": {
+    "scaleTypeSelect": {
+      "title": "Type",
+      "enum": (vlSchema as any).definitions.ScaleType.enum,
+      "type": "string"
+    },
+  },
+};
+
+export const AXIS_TITLE_UISCHEMA = {
```
kanitw: Why are XXX_UISCHEMA declared separately from XXX_SCHEMA?

kanitw: I also would like to make sure all these constants have proper interfaces so they are consistently declared.

src/components/encoding-pane/property-editor.tsx
```diff
@@ -14,24 +19,61 @@ export interface PropertyEditorProps extends ActionHandler<SpecEncodingAction> {
   fieldDef: ShelfFieldDef;
 }
 
+export interface PropertyEditorSchema {
+  uiSchema: any;
+  schema: any;
+}
+
 export class PropertyEditorBase extends React.PureComponent<PropertyEditorProps, {}> {
+  private static getPropertyEditorSchema(prop: string, nestedProp: string): PropertyEditorSchema {
```
kanitw: Do they use anything from this class, if not, better just declare in the schema file so you only export this function and everything else remain encapsulate in the file?

src/components/encoding-pane/property-editor.tsx
```diff
@@ -14,24 +19,61 @@ export interface PropertyEditorProps extends ActionHandler<SpecEncodingAction> {
...
+    if (prop === 'scale') {
+      if (nestedProp === 'type') {
+        return {
+          schema: SCALE_TYPE_SCHEMA,
```
kanitw: I think `SCALE_TYPE_SCHEMA` is actually used only once.  We can stick the return value in here, esp. if we move this function to the other file already?

src/components/encoding-pane/property-editorSchema.ts
```diff
@@ -0,0 +1,79 @@
...
+  "axisTitle": {
+    "ui:autofocus": true,
+    "ui:emptyValue": ""
+  }
+};
+
+export const AXIS_ORIENT_UISCHEMA = {
+  "orient": {
+    "ui:widget": "select",
```
kanitw: UI SCHEMA for these last 3 items are exactly the same.  

Better extract a well-named common `const`

src/components/encoding-pane/property-editor.tsx
```diff
@@ -14,24 +19,61 @@ export interface PropertyEditorProps extends ActionHandler<SpecEncodingAction> {
...
   public render() {
-    const {prop, nestedProp, fieldDef} = this.props;
-    const uiSchema = {
-      "ui:title": `${prop} ${nestedProp}`,
-      "ui:placeholder": "auto",
-      "ui:emptyValue": "auto"
-    };
-    const formData = fieldDef.scale ? fieldDef.scale.type : 'auto';
+    const {prop, nestedProp} = this.props;
+    const schemaObj = PropertyEditorBase.getPropertyEditorSchema(prop, nestedProp);
```
kanitw: const {schema, uiSchema} =

src/components/encoding-pane/property-editorSchema.ts
kanitw: Plus, I see that the key in the SCHEMA and UISCHEMA are always the same, but given you have to type them separately, this can be error prone. 

Better declare a factory that takes: 
1) parent title 
2) property name (if there is always only one property for each declared schema) 
3) schema (title, enum, type)
4) uiSchema.  

and consistently generate it. 


or even better replace property name with nestedProp and always generate a consistent name pattern.

src/components/encoding-pane/property-editorSchema.ts
kanitw: This answer the first question, but not 

> Why is it called "orient", not "axisOrient?

(Tip: use `>` to highlight a question u answer when reply a comment with multiple questions.)

ssharif6: Ah thanks for the tip, did not know that :)

> Why is it called "orient", not "axisOrient"

Good question. The new schema convention will have the properties key be `prop_nestedProp` (or do you prefer just `propNestedProp`

kanitw: `prop_nestedProp` is way better (simpler to generate, easy to split)

kanitw: `propNestedProp` is more JS style though, but l still think `_` is better

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,96 @@
...
+  return s.charAt(0).toUpperCase() + s.slice(1);
+}
+
+/*
+*  NOTE: factory method where propertyKey follows naming convention: Capitalized Prop + Capitalized NestedProp
+*  Example: {prop: axis, nestedProp: orient} translates to "AxisOrient"
+*/
+function generateSchema(prop: string, nestedProp: string, uiSchema: any, schemaEnum?: any): PropertyEditorSchema {
+  prop = capitalizeFirstLetter(prop);
+  nestedProp = nestedProp ? capitalizeFirstLetter(nestedProp) : nestedProp;
```
kanitw: We agreed to use `_`, not camelCase nor PascalCase.  

https://github.com/vega/voyager/pull/789#discussion_r183281143

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface PropertyEditorSchema {
+  uiSchema: any;
+  schema: PropertyEditorFormatSchema;
+}
+
+export function getPropertyEditorSchema(prop: string, nestedProp: string): PropertyEditorSchema {
+  if (prop === 'scale') {
+    if (nestedProp === 'type') {
+      return generateSchema(prop, nestedProp, SCALE_TYPE_UISCHEMA,
```
kanitw: 1) Why declaring UISCHEMA below, not up here.

2) UISCHEMA shouldn't contain the key they you automatically generate anymore.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,96 @@
+import * as vlSchema from 'vega-lite/build/vega-lite-schema.json';
+
+export interface PropertyEditorFormatSchema {
+  type: string;
+  title: string;
+  properties: any;
+}
+
+export interface PropertyEditorSchema {
+  uiSchema: any;
```
kanitw: Is it really any?  Does the lib that we should contains some schema?

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,96 @@
+import * as vlSchema from 'vega-lite/build/vega-lite-schema.json';
+
+export interface PropertyEditorFormatSchema {
+  type: string;
+  title: string;
+  properties: any;
```
kanitw: This is a JSON schema, so it's not really `any`.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,96 @@
...
+};
+
+// Reuse for any element with Input Text element
+const TEXT_UISCHEMA = {
+  "ui:autofocus": true,
+  "ui:emptyValue": ""
+};
+
+const AXIS_TITLE_UISCHEMA = {
+  "AxisTitle": TEXT_UISCHEMA
```
kanitw: The keys (e.g., `AxisTitle`) can be auto generate in your factory method, why typing manually for each of these?

kanitw: Let's make it clear that I don't want to see any  (`<PROP_NESTEDPROP>_UISCHEMA`) constants left.

src/components/encoding-pane/property-editor.tsx
```diff
@@ -45,16 +44,37 @@ export class PropertyEditorBase extends React.PureComponent<PropertyEditorProps,
...
+          value
+        }
+      });
+    }
+  }
+
+  private getFormData() {
+    const {fieldDef} = this.props;
+    return {
+      'ScaleType': fieldDef.scale ? fieldDef.scale.type : undefined,
```
kanitw: 1) This is still manually typed, which is tedious and error prone.
2) I believe this will depend on field type, just like `getFieldPropertyGroupIndex()`.

Since you have all the information needed from [`getFieldPropertyGroupIndex()`]( https://github.com/vega/voyager/pull/789/files#diff-6e0fc12971841df1d666cfddd2fb434dR139),  I think you can extract `getFieldPropertyGroupIndex()` to be a part of the schema file and use it both here and in `field-customizer.tsx`.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,96 @@
...
+*  Example: {prop: axis, nestedProp: orient} translates to "AxisOrient"
+*/
+function generateSchema(prop: string, nestedProp: string, uiSchema: any, schemaEnum?: any): PropertyEditorSchema {
+  prop = capitalizeFirstLetter(prop);
+  nestedProp = nestedProp ? capitalizeFirstLetter(nestedProp) : nestedProp;
+  const propertyKey = nestedProp ? prop + nestedProp : prop;
+  const schema: PropertyEditorFormatSchema = {
+    title: prop,
+    type: "object",
+    properties: {}
```
kanitw:
```
properties: {
   [propertyKey]: {
     "type": "string",
     "title": nestedProp || prop,
      ...(schemaEnum ? {enum: schemaEnum} : {}) 
   }
}
```

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+  return {
+    schema: schema,
+    uiSchema: uiSchema
+  };
+}
+
+// NOTE: Each Nested Property requires its own UISCHEMA object
+
+// Reuse for any UISCHEMA with Select element
```
kanitw: Move reused constants to the top of the file.  

Good practice is to declare the constants before using them.

.gitignore
```diff
@@ -5,3 +5,4 @@ node_modules
 coverage
 npm-debug.log
 .idea
+package-lock.json
```
kanitw: Can we not include package-lock.json?

src/components/encoding-pane/field-customizer.tsx
```diff
@@ -1,47 +1,70 @@
...
+              keys.map((encodingType, i) => {
+                const customProps: CustomProp[] = propertyGroupIndex[encodingType];
+                return (
+                  <TabPanel key={encodingType + i}>
+                    {
+                      customProps.map(customizableProp => {
+                        const {prop, nestedProp} = customizableProp;
+                        return (
+                          <PropertyEditor
+                            key={JSON.stringify({prop, nestedProp})}
```
kanitw: concat with underscore is cheaper than stringify?

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,195 @@
...
+
+export function getPropertyEditorSchema(prop: string, nestedProp: string, propTab: string): PropertyEditorSchema {
+  const title = generateTitle(prop, nestedProp, propTab);
+  const baseSchemaProp = {
+    title: title
+  };
+  if (prop === 'scale') {
+    if (nestedProp === 'type') {
+      const schemaProperty: StringSchema = {
+        ...baseSchemaProp as StringSchema,
```
kanitw: Why using `as`?  looks like a hack. 

Why having baseSchemaProp at all anyway?   (Why not adding `title` directly to each of these cases?)

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,195 @@
...
+    type: 'object',
+    properties: {
+      [propertyKey]: schemaProp
+    }
+  };
+  const uiSchema: UISchema = {
+    [propertyKey]: uiSchemaItem
+  };
+
+  return {
```
kanitw: return {schema, uiSchema};

.gitignore
```diff
@@ -5,4 +5,3 @@ node_modules
 coverage
 npm-debug.log
 .idea
-package-lock.json
```
kanitw: My  bad, you commit the right thing ><" sorry!

kanitw:
- [ ] Cannot click outside the popup to hide the popup (Try the function popup). 

- [ ] The height for the tab is too much, even bigger than the pill/capsule for each encoding on the shelf. 

- [ ] Font size, padding inconsistent with the function popup (The font size should be the same. The padding is very big.)

ssharif6:
ClickOutside function implemented + more compact UI:
fieldcustomizerupdatedui

PR #790, opened 2018-04-07
Add hide related-views config to lib-voyager

ssharif6:
Issue #786 

Hiding Related-Views is now connected to VoyagerConfig. This allows clients to toggle whether or not to have related-views by default in the Voyager Config.


src/models/config.ts
```diff
@@ -4,12 +4,14 @@ export interface VoyagerConfig {
...
 };
 
 export const DEFAULT_VOYAGER_CONFIG: VoyagerConfig = {
   showDataSourceSelector: true,
   serverUrl: null,
   manualSpecificationOnly: false,
   hideHeader: false,
-  hideFooter: false
+  hideFooter: false,
+  hideRelatedViews: false
```
kanitw: You add it here, but how does it get applied when people set it?

src/components/view-pane/index.tsx
```diff
@@ -69,27 +69,28 @@ class ViewPaneBase extends React.PureComponent<ViewPaneProps, {}> {
...
-    const { manualSpecificationOnly } = this.props.config;
+    const { manualSpecificationOnly, hideRelatedViews } = this.props.config;
     const relatedViews = !manualSpecificationOnly && (
-      <div className="pane" styleName={!this.props.relatedViews.isHidden ?
-      "view-pane-related-views" : "view-pane-related-views-hide"}>
-        <RelatedViewsButton
-          relatedViews={this.props.relatedViews}
-          handleAction={this.props.handleAction}
-        />
+      <div className="pane" styleName={this.props.relatedViews.isHidden === undefined ? hideRelatedViews ?
```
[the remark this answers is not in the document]
kanitw: Plus, you reuse the condition for `relatedViews.isHidden` and `hideRelatedViews` rules across this file.  Why don't you just declare `hideRelatedViews` that combines the two once and use it everywhere?

src/components/view-pane/related-views-button.tsx
```diff
@@ -18,25 +22,39 @@ export class RelatedViewsButtonBase extends React.PureComponent<RelatedViewsButt
...
 }
 
-export const RelatedViewsButton = (CSSModules(RelatedViewsButtonBase, styles));
+
+export const RelatedViewsButton = connect(
+  (state: State) => {
+    return {
+      config: selectConfig(state),
+      relatedViews: selectRelatedViews(state)
+    } as RelatedViewsButtonProps;
```
kanitw: Why casting here?

src/reducers/related-views.ts
```diff
@@ -4,11 +4,11 @@ import { DEFAULT_RELATED_VIEWS, RelatedViews } from '../models/related-views';
 
 export function relatedViewsReducer(
   relatedViewToggler: RelatedViews = DEFAULT_RELATED_VIEWS, action: Action): RelatedViews {
-  const { isHidden } = relatedViewToggler;
   switch (action.type) {
     case RELATED_VIEWS_HIDE_TOGGLE: {
+      const {hideRelatedViews} = action.payload;
```
kanitw: `hideRelatedViews` here is "current" status of hideRelatedViews that you want to switch form.  The current name doesn't make it clear whether it is an old value (that you don't want) or a new value that you want.

I'd say since this is an action, better make rename `hideRelatedViews` and make it `newIsHidden` state (not `currentIsHidden`).

src/reducers/related-views.ts
```diff
@@ -4,11 +4,11 @@ import { DEFAULT_RELATED_VIEWS, RelatedViews } from '../models/related-views';
 
 export function relatedViewsReducer(
   relatedViewToggler: RelatedViews = DEFAULT_RELATED_VIEWS, action: Action): RelatedViews {
-  const { isHidden } = relatedViewToggler;
   switch (action.type) {
     case RELATED_VIEWS_HIDE_TOGGLE: {
+      const newIsHidden = action.payload.hideRelatedViews;
```
kanitw: Rename the property in the payload too, otherwise, there is no point renaming it locally here.

src/reducers/related-views.ts
```diff
@@ -4,11 +4,11 @@ import { DEFAULT_RELATED_VIEWS, RelatedViews } from '../models/related-views';
...
       return {
-        isHidden: !isHidden
+        isHidden: !newIsHidden
```
kanitw: This is incorrect.  Why do you set isHidden to negation of `newIsHidden`?

src/components/view-pane/related-views-button.tsx
```diff
@@ -18,25 +22,38 @@ export class RelatedViewsButtonBase extends React.PureComponent<RelatedViewsButt
   }
 
   public render() {
+    const {config, relatedViews} = this.props;
+    const {hideRelatedViews} = config;
+    const hideCondition = (relatedViews.isHidden === undefined && hideRelatedViews) ||
```
kanitw: You still have redundant condition here, below for `onHideClick()`, and in the other file.   

Also, not sure why you have to write the condition in two different ways, although they should be the same. 

Since this is reused across file, why not consolidate the logic?

For example, you could just remove `relatedViews` and `config` from `RelatedViewsButtonProps`  and just have `hideRelatedViews` that you pass in (and `hideRelatedViews` should be already a combined version of what you have in relatedViews state and config state). 

Basically, related views button should be relatively dumb -- it doesn't have to know that externally you combine two properties for get the value whether you know related views or not.  

(See https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0, if you haven't read about it.)

src/components/view-pane/related-views-button.tsx
```diff
@@ -18,25 +22,38 @@ export class RelatedViewsButtonBase extends React.PureComponent<RelatedViewsButt
...
+      (relatedViews.isHidden !== undefined && relatedViews.isHidden);
     return (
       <div styleName="right">
         <a onClick={this.onHideClick}>
-          {this.props.relatedViews.isHidden ? 'Show' : 'Hide'}
+          {hideCondition ? "Show" : "Hide"}
```
kanitw: for JS parts, use single quotes -- not sure why you have to change to double quotes.

src/components/view-pane/related-views-button.tsx
```diff
@@ -18,25 +22,38 @@ export class RelatedViewsButtonBase extends React.PureComponent<RelatedViewsButt
...
+      type: RELATED_VIEWS_HIDE_TOGGLE,
+      payload: {
+        hideRelatedViews: relatedViews.isHidden === undefined ? config.hideRelatedViews : relatedViews.isHidden
+      }
     });
   }
 }
 
-export const RelatedViewsButton = (CSSModules(RelatedViewsButtonBase, styles));
+export const RelatedViewsButton = connect(
```
kanitw: This "connect" is unnecessary.  (Better change the properties to have the "hide" state like commented above, that's it.)

src/components/view-pane/index.tsx
```diff
@@ -68,32 +68,29 @@ class ViewPaneBase extends React.PureComponent<ViewPaneProps, {}> {
...
+    const {isQuerySpecific, relatedViews} = this.props;
+    const {manualSpecificationOnly, hideRelatedViews} = this.props.config;
+    const hideCondition = (relatedViews.isHidden === undefined && hideRelatedViews) ||
+      (relatedViews.isHidden !== undefined && relatedViews.isHidden);
+    const relatedViewsElement = !manualSpecificationOnly && (
+      <div className="pane" styleName={hideCondition ? "view-pane-related-views-hide" : "view-pane-related-views"}>
         <RelatedViewsButton
-          relatedViews={this.props.relatedViews}
+          hideRelatedViews={hideCondition}
           handleAction={this.props.handleAction}
```
kanitw: This is inconsistent -- why not putting handleAction above in `const {isQuerySpecific, relatedViews} = this.props;`

src/models/config.ts
```diff
@@ -4,12 +4,14 @@ export interface VoyagerConfig {
...
 };
 
 export const DEFAULT_VOYAGER_CONFIG: VoyagerConfig = {
   showDataSourceSelector: true,
   serverUrl: null,
   manualSpecificationOnly: false,
   hideHeader: false,
-  hideFooter: false
+  hideFooter: false,
+  hideRelatedViews: true
```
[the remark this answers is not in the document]
ssharif6: Reverted back to false

PR #792, opened 2018-04-11
Add wildcards property to VoyagerConfig

ssharif6:
Refactor VoyagerConfig to have property `wildcards` with possible values `enabled|disabled` and property `relatedViews` with possible values `initiallyCollapsed | initiallyShown | disabled`

Case 1: `relatedViews: disabled`, `wildcards: enabled`
image

Case 2: `relatedViews: initiallyCollapsed`, `wildcards: enabled`
image

Case 3: `relatedViews: initiallyShown`, `wildcards: disabled`
image

### DIFF
--- a/src/components/data-pane/index.tsx
+++ b/src/components/data-pane/index.tsx
@@ -17,7 +17,7 @@ export class DataPaneBase extends React.PureComponent<DataPanelProps, {}> {
   public render() {
     const {name} = this.props.data;
     const fieldCount = this.props.data.schema.fieldSchemas.length;
-    const {showDataSourceSelector, relatedViews} = this.props.config;
+    const {showDataSourceSelector, wildcards} = this.props.config;
 
     const fields = fieldCount > 0 ? (
       <div styleName="data-pane-section">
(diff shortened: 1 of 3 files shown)


src/components/view-pane/index.tsx
```diff
@@ -73,7 +73,7 @@ class ViewPaneBase extends React.PureComponent<ViewPaneProps, {}> {
     const collapseRelatedViews = relatedViews.isCollapsed === undefined ? config.relatedViews === 'initiallyCollapsed' :
       relatedViews.isCollapsed;
 
-    const relatedViewsElement = config.relatedViews !== 'disabled' && (
+    const relatedViewsElement = config.wildcards !== 'disabled' && (
```
felixcodes: I don't think this was supposed to change

src/components/view-pane/index.tsx
```diff
@@ -73,7 +73,7 @@ class ViewPaneBase extends React.PureComponent<ViewPaneProps, {}> {
     const collapseRelatedViews = relatedViews.isCollapsed === undefined ? config.relatedViews === 'initiallyCollapsed' :
       relatedViews.isCollapsed;
 
-    const relatedViewsElement = config.relatedViews !== 'disabled' && (
```
felixcodes: This shouldn't change at all.

felixcodes:
cc: @ZeningQu - When this PR is merged and released you'll use the `relatedViews: disabled` and `wildcards: disabled` configs. `wildcards: disabled` refers strictly to wildcard fields -- so you'll still be able to use auto marks.

PR #794, opened 2018-04-15
Fix closing customizer bug

### LINKED ISSUE #775: Bug: Can't close the position drop-down pop-up in the encoding panel (for x and y)

ssharif6:
Fix #775 
Only allow property-editor pane to be opened when `fieldDef` is defined. 
customizerbuffix


kanitw:
I think you mean: 

> Only allow property-editor pane to be opened when fieldDef is defined.

Only allow property-editor pane to "remain" opened when fieldDef is defined.

PR #799, opened 2018-05-01
More Custom Encoding

ssharif6:
Extension of PR #789  and part of Issue #757

* Add support for Positional & Mark property channels (except text and detail, as `format` prop is not supported in Compass yet).
* Click outside of form will close field-customizer

Schema design & gifs for each mark encoding found here (for mark channels): https://docs.google.com/document/d/1n4pd6q3jMbZ_BYoSLYsMzTV9NOc25ZTJZ8oAGav3zP4/edit?usp=sharing

### DIFF
--- a/src/components/encoding-pane/encoding-shelf.tsx
+++ b/src/components/encoding-pane/encoding-shelf.tsx
@@ -86,7 +86,7 @@ class EncodingShelfBase extends React.PureComponent<
             attachment="top left"
             targetAttachment="bottom left"
           >
-            {(fieldDef && !isWildcardChannelId(id) && (id.channel === 'x' || id.channel === 'y')) ?
+            {(fieldDef && !isWildcardChannelId(id)) ?
               <span onClick={this.toggleCustomizer} ref={this.fieldHandler}>
                 {channelName}{' '} <i className={'fa fa-caret-down'}/>
               </span> :
@@ -168,7 +168,7 @@ class EncodingShelfBase extends React.PureComponent<
 
   protected onRemove() {
     const {id, handleAction} = this.props;
-    this.toggleCustomizer();
+    this.closePopup();
     handleAction({
       type: SPEC_FIELD_REMOVE,
       payload: id
@@ -219,7 +219,8 @@ class EncodingShelfBase extends React.PureComponent<
   }
 
   private handleClickOutside(e: any) {
-    if (this.fieldCustomizer && (this.fieldCustomizer.contains(e.target) || this.encodingShelf.contains(e.target))) {
+    if (this.fieldCustomizer && this.encodingShelf && (this.fieldCustomizer.contains(e.target) ||
+      this.encodingShelf.contains(e.target))) {
       return;
     }
     this.closePopup();
--- a/src/components/encoding-pane/field-customizer.tsx
+++ b/src/components/encoding-pane/field-customizer.tsx
@@ -22,46 +22,48 @@ export class FieldCustomizerBase extends React.PureComponent<FieldCustomizerProp
   public render() {
     const {shelfId, handleAction, fieldDef} = this.props;
     const propertyGroupIndex = getFieldPropertyGroupIndex(shelfId, fieldDef);
-    const keys = Object.keys(propertyGroupIndex);
+    const keys = propertyGroupIndex ? Object.keys(propertyGroupIndex) : undefined;
     return (
       <div styleName='field-customizer'>
-        <Tabs>
-          <TabList>
-            {
-              keys.map((encodingType, i) => {
-                return (
-                  <Tab key={i}>{encodingType}</Tab>
-                );
-              })
-            }
-          </TabList>
-          <div>
-            {
-              keys.map((encodingType, i) => {
-                const customProps: CustomProp[] = propertyGroupIndex[encodingType];
-                return (
-                  <TabPanel key={encodingType + i}>
-                    {
-                      customProps.map(customizableProp => {
-                        const {prop, nestedProp} = customizableProp;
-                        return (
-                          <PropertyEditor
-                            key={prop + '_' + nestedProp}
-                            prop={prop}
-                            nestedProp={nestedProp}
-                            shelfId={shelfId}
-                            fieldDef={fieldDef}
-                            handleAction={handleAction}
-                            propTab={encodingType}
-                          />
-                        );
-                      })
-                    }
-                  </TabPanel>
-                );
-              })}
-          </div>
-        </Tabs>
+        {keys && (
+          <Tabs>
+            <TabList>
+              {
+                keys.map((encodingType, i) => {
+                  return (
+                    <Tab key={'tab' + i}>{encodingType}</Tab>
+                  );
+                })
+              }
+            </TabList>
+            <div>
+              {
+                keys.map((encodingType, i) => {
+                  const customProps: CustomProp[] = propertyGroupIndex[encodingType];
+                  return (
+                    <TabPanel key={'tabPanel' + encodingType + i}>
+                      {
+                        customProps.map(customizableProp => {
+                          const {prop, nestedProp} = customizableProp;
+                          return (
+                            <PropertyEditor
+                              key={prop + '_' + nestedProp + i}
+                              prop={prop}
+                              nestedProp={nestedProp}
+                              shelfId={shelfId}
+                              fieldDef={fieldDef}
+                              handleAction={handleAction}
+                              propTab={encodingType}
+                            />
+                          );
+                        })
+                      }
+                    </TabPanel>
+                  );
+                })}
+            </div>
+          </Tabs>
+        )}
       </div>
     );
   }
--- a/src/components/encoding-pane/property-editor-schema.ts
+++ b/src/components/encoding-pane/property-editor-schema.ts
@@ -1,8 +1,12 @@
 import {ExpandedType} from 'compassql/build/src/query/expandedtype';
 import {Channel} from 'vega-lite/build/src/channel';
+import {contains} from 'vega-lite/build/src/util';
 import * as vlSchema from 'vega-lite/build/vega-lite-schema.json';
 import {ShelfFieldDef, ShelfId} from '../../models/shelf/spec';
 
+// ------------------------------------------------------------------------------
+// Schema Interfaces
+// ------------------------------------------------------------------------------
 export interface SchemaProperties {
   [key: string]: SchemaProperty;
 }
@@ -12,17 +16,27 @@ export interface ObjectSchema {
   title?: string;
   properties: SchemaProperties;
 }
+
 export interface StringSchema {
   type: 'string';
   title: string;
   enum?: string[];
+  default?: string;
+}
+
+export interface IntegerSchema {
+  type: 'integer';
+  title: string;
+  minimum: number;
+  maximum: number;
+  multipleOf: number;
 }
 
 export function isStringSchema(schema: SchemaProperty): schema is StringSchema {
   return schema.type === 'string';
 }
 
-export type SchemaProperty = ObjectSchema | StringSchema;
+export type SchemaProperty = ObjectSchema | StringSchema | IntegerSchema;
 
 export interface UISchema {
   [key: string]: UISchemaItem;
@@ -31,7 +45,6 @@ export interface UISchema {
 // NOTE: keys for these interfaces follow the requirement for react-jsonSchema form
 // (https://mozilla-services.github.io/react-jsonschema-form/)
 export interface UISchemaItem {
-  'ui:autofocus'?: boolean;
   'ui:widget'?: string;
   'ui:placeholder'?: string;
   'ui:emptyValue'?: string;
@@ -42,9 +55,21 @@ export interface PropertyEditorSchema {
   schema: ObjectSchema;
 }
 
+// ------------------------------------------------------------------------------
+// Default UISchema objects for react-jsonschema-form
+// ------------------------------------------------------------------------------
 const DEFAULT_TEXT_UISCHEMA: UISchemaItem = {
-  'ui:autofocus': true,
   'ui:emptyValue': ''
+};
+
+const DEFAULT_DISCRETE_SCALE_ARRAY_UISCHEMA: UISchemaItem = {
+  ...DEFAULT_TEXT_UISCHEMA,
+  'ui:placeholder': 'a, b, c...'
+};
+
+const DEFAULT_CONTINUOUS_SCALE_ARRAY_UISCHEMA: UISchemaItem = {
+  ...DEFAULT_TEXT_UISCHEMA,
+  'ui:placeholder': 'Min, Max'
 };
 
 const DEFAULT_SELECT_UISCHEMA: UISchemaItem = {
@@ -53,6 +78,10 @@ const DEFAULT_SELECT_UISCHEMA: UISchemaItem = {
   'ui:emptyValue': 'auto'
 };
 
+// ------------------------------------------------------------------------------
+// Channel-Field Indexes for custom encoding
+// Key is Tab name, value is list of fieldDef properties
+// ------------------------------------------------------------------------------
 const POSITION_FIELD_QUANTITATIVE_INDEX = {
   'Common': [
     {
@@ -67,8 +96,7 @@ const POSITION_FIELD_QUANTITATIVE_INDEX = {
       prop: 'stack'
     }
   ],
-  'Scale': ['type'].map(p => ({prop: 'scale', nestedProp: p})),
-  'Axis': ['orient', 'title'].map(p => ({prop: 'axis', nestedProp: p}))
+  'Axis': ['orient', 'title'].map(p => ({prop: 'axis', nestedProp: p})),
 };
 
 const POSITION_FIELD_NOMINAL_INDEX = {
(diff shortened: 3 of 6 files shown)


src/components/encoding-pane/field-customizer.tsx
```diff
@@ -12,35 +14,141 @@ export interface FieldCustomizerProps extends ActionHandler<SpecEncodingAction>
...
+      nestedProp: 'orient'
+    },
+    {
+      prop: 'axis',
+      nestedProp: 'title'
+    }
+  ]
+};
+
+const POSITION_FIELD_TEMPORAL_INDEX = {
```
kanitw: This seems to be reverted ...

kanitw:
I'm changing merge target here to the new feature branch.  Please drop the first commit to avoid conflicts.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -42,17 +55,33 @@ export interface PropertyEditorSchema {
...
+// Default UISchema objects for react-jsonschema-form
+// ------------------------------------------------------------------------------
 const DEFAULT_TEXT_UISCHEMA: UISchemaItem = {
-  'ui:autofocus': true,
   'ui:emptyValue': ''
 };
 
+const DEFAULT_DISCRETE_SCALE_ARRAY_UISCHEMA: UISchemaItem = {
+  ...DEFAULT_TEXT_UISCHEMA,
+  'ui:placeholder': 'a, b, c...'
```
kanitw: c, ...

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -42,17 +55,33 @@ export interface PropertyEditorSchema {
...
+};
+
+const DEFAULT_CONTINUOUS_SCALE_ARRAY_UISCHEMA: UISchemaItem = {
+  ...DEFAULT_TEXT_UISCHEMA,
+  'ui:placeholder': 'Min, Max'
```
kanitw: This won't be clear to users that min is number (same for max)

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -81,34 +109,65 @@ const POSITION_FIELD_TEMPORAL_INDEX = {
...
 
-function generateTitle(prop: string, nestedProp: string, propTab: string): string {
-  let title;
-  if (propTab === 'Common') {
-    title = nestedProp ? prop + ' ' + nestedProp : prop;
-  } else {
-    title = nestedProp || prop;
-  }
+const SIZE_CHANNEL_FIELD_INDEX = {
+  'Legend': ['orient', 'title', 'type'].map(p => ({prop: 'legend', nestedProp: p})),
```
kanitw: why does legend type matter for size?

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -81,34 +109,65 @@ const POSITION_FIELD_TEMPORAL_INDEX = {
   'Axis': ['orient', 'title'].map(p => ({prop: 'axis', nestedProp: p}))
 };
 
-// Capitalize first letter for aesthetic purposes in form
-function toTitleCase(str: string) {
-  return str.replace(/\w\S*/g, txt => {
-    return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();
-  });
-}
+const COLOR_CHANNEL_FIELD_INDEX = {
```
kanitw: Why don't these channel have "common" tab like position?

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -181,9 +279,59 @@ export function generateFormData(shelfId: ShelfId, fieldDef: ShelfFieldDef) {
...
+      const fData = fieldDef[prop] ? nestedProp ? fieldDef[prop][nestedProp] : fieldDef[prop] : undefined;
+      // handle empty input for text input
+      formData[propertyKey] = fData === '?' ? '' : fData;
     }
   }
 
   return formData;
 }
+
+function filterEnumElements(fieldDef: ShelfFieldDef, prop: string, nestedProp: string, possibleVals: string[]) {
```
kanitw: This method is very convoluted.

src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -168,7 +168,7 @@ class EncodingShelfBase extends React.PureComponent<
 
   protected onRemove() {
     const {id, handleAction} = this.props;
-    this.toggleCustomizer();
+    this.closePopup();
```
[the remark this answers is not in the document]
ssharif6: This was needed to allow the field-customizer UI to close when clicking outside of the customizer.

src/components/encoding-pane/property-editor-schema.ts
ssharif6: Wasn't sure which ones would be "most common". @felixcodes any ideas?

src/components/encoding-pane/property-editor-schema.ts
ssharif6: I figured having a generic helper method filter enum values based on the fieldDef type would be helpful 

The intended purpose of this method is to filter enum values based on the `fieldDef`'s type. So for instance, only appropriate scale types are displayed, or appropriate color schemes based on the definition of the `fieldDef`.

src/components/encoding-pane/property-editor-schema.ts
ssharif6: good point

src/components/encoding-pane/property-editor-schema.ts
kanitw: > I could do this in generatePropertyEditorSchema method as well if having a separate method isn't needed.

Yep, that's exactly what you should do.  In general you want to minimize code jump when reading the code, unless you have some common methods that are repetitive.

In this case, they are not truly repetitive, but you made them the same function.  So this basically add unnecessary complexity in two ways: (1)  unnecessary code jump and (2)more conditional logic to distinguish things that are not supposed to be in the same method.

I could agree that another good reason for a code jump is when a method is too long, but I still disagree with this way of method extraction.

src/components/encoding-pane/encoding-shelf.tsx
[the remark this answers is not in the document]
kanitw: Ok but it's irrelevant for the purpose of this PR, thus making this commit not atomic. 

(No worry about it at this point since I already get it, but to simplify my review, please make your commits atomic.)

src/components/encoding-pane/property-editor-schema.ts
kanitw: Legend title, scale range or scheme

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -81,34 +109,85 @@ const POSITION_FIELD_TEMPORAL_INDEX = {
...
+      nestedProp: 'type'
+    }
+  ],
+  'Legend': ['orient', 'title'].map(p => ({prop: 'legend', nestedProp: p})),
+  'Scale': ['domain', 'range', 'type'].map(p => ({prop: 'scale', nestedProp: p}))
+};
 
-  return toTitleCase(title);
-}
+const SHAPE_CHANNEL_FIELD_INDEX = {
```
ssharif6: Not sure if a Common tab is needed here, but if it does it's easy to add in.

kanitw: If you gonna have two tabs with two properties each, better just have one tab with 4 properties.  Same for other channels :)

ssharif6: Sounds good. This means each positional channel will only have one common tab, since there are only 4 properties (for right now, will add more later).

image


Is this ok?

kanitw: I guess it is ok.  The current state is a bit funny since 1 tab = better have no tab too.

But we will add more pretty soon, so let's just keep the tab. :)

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -42,17 +55,33 @@ export interface PropertyEditorSchema {
...
 };
 
+const DEFAULT_DISCRETE_SCALE_ARRAY_UISCHEMA: UISchemaItem = {
+  ...DEFAULT_TEXT_UISCHEMA,
+  'ui:placeholder': 'a, b, c, ...'
+};
+
+const DEFAULT_CONTINUOUS_SCALE_ARRAY_UISCHEMA: UISchemaItem = {
+  ...DEFAULT_TEXT_UISCHEMA,
+  'ui:placeholder': 'Min Num, Max Num'
```
kanitw: "Number"

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -132,12 +262,45 @@ export function getPropertyEditorSchema(prop: string, nestedProp: string, propTa
       type: 'string'
     };
     return generateSchema(prop, nestedProp, DEFAULT_SELECT_UISCHEMA, propTab, schemaProperty);
+  } else if (prop === 'size') {
+    const schemaProperty: StringSchema = {
```
kanitw: Pretty much almost all cases of this method need to type 

```
{
  title, 
  type: string
}
```

I wonder if there is a way to reduce reduntant typing required.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -181,9 +350,41 @@ export function generateFormData(shelfId: ShelfId, fieldDef: ShelfFieldDef) {
       const prop = customProp.prop;
       const nestedProp = customProp.nestedProp;
       const propertyKey = nestedProp ? prop + '_' + nestedProp : prop;
-      formData[propertyKey] = fieldDef[prop] ? nestedProp ? fieldDef[prop][nestedProp] : fieldDef[prop] : undefined;
+      const fData = fieldDef[prop] ? nestedProp ? fieldDef[prop][nestedProp] : fieldDef[prop] : undefined;
+      // Display empty string when '?' is passed in to retrieve auto generated visualization
```
kanitw: auto-generated visualization?    Do you mean default value.  

Also, when would '?' be passed in?  Can you explain more?

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -63,81 +92,217 @@ const POSITION_FIELD_QUANTITATIVE_INDEX = {
...
 
 
+// TODO: Eventually refactor to Vega-Lite
+function getSupportedScaleTypes(shelfId: ShelfId, fieldDef: ShelfFieldDef): string[] {
+  let scaleTypes: string[] = [];
+  if (contains([Channel.X, Channel.Y], shelfId.channel)) {
+    if (fieldDef.type === ExpandedType.QUANTITATIVE) {
+      scaleTypes = ["linear", "log", "pow", "sqrt"];
+    } else if (fieldDef.type === ExpandedType.TEMPORAL) {
+      scaleTypes = ["bin-linear", "time", "utc", "ordinal", "bin-ordinal", "point", "band"];
```
kanitw: ["time", "utc"]

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -63,81 +92,217 @@ const POSITION_FIELD_QUANTITATIVE_INDEX = {
...
+      ...baseUiSchema
+    };
+    return generateSchema(prop, nestedProp, DEFAULT_TEXT_UISCHEMA, propTab, schemaProperty);
   } else {
     throw new Error('Property combination not recognized');
   }
 }
 
 
+// TODO: Eventually refactor to Vega-Lite
```
kanitw: As mentioned before, please file an issue.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -63,81 +92,217 @@ const POSITION_FIELD_QUANTITATIVE_INDEX = {
...
+    } else if (fieldDef.type === ExpandedType.TEMPORAL) {
+      scaleTypes = ["bin-linear", "time", "utc", "ordinal", "bin-ordinal", "point", "band"];
+    } else if (fieldDef.type === ExpandedType.NOMINAL || fieldDef.type === ExpandedType.ORDINAL) {
+      scaleTypes = ["point", "band"];
+    }
+  } else if (shelfId.channel === Channel.COLOR) {
+    if (fieldDef.type === ExpandedType.QUANTITATIVE) {
+      scaleTypes = ["linear", "pow", "sqrt", "log", "sequential"];
+    } else if (fieldDef.type === ExpandedType.NOMINAL || fieldDef.type === ExpandedType.ORDINAL) {
+      scaleTypes = ["ordinal", "point"];
```
kanitw: why point?  -- ["ordinal"]

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -63,81 +92,217 @@ const POSITION_FIELD_QUANTITATIVE_INDEX = {
...
+    return generateSchema(prop, nestedProp, DEFAULT_TEXT_UISCHEMA, propTab, schemaProperty);
   } else {
     throw new Error('Property combination not recognized');
   }
 }
 
 
+// TODO: Eventually refactor to Vega-Lite
+function getSupportedScaleTypes(shelfId: ShelfId, fieldDef: ShelfFieldDef): string[] {
+  let scaleTypes: string[] = [];
```
kanitw: Why do you need to `let`.  Better just return them directly.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -63,81 +92,217 @@ const POSITION_FIELD_QUANTITATIVE_INDEX = {
...
+    } else if (fieldDef.type === ExpandedType.TEMPORAL) {
+      scaleTypes = ["time", "utc", "sequential"];
+    }
+  } else if (shelfId.channel === Channel.SIZE) {
+    if (fieldDef.type === ExpandedType.QUANTITATIVE) {
+      scaleTypes = ["linear", "pow", "sqrt", "log"];
+    } else if (fieldDef.type === ExpandedType.NOMINAL || fieldDef.type === ExpandedType.ORDINAL) {
+      scaleTypes = ["point", "band"];
+    } else if (fieldDef.type === ExpandedType.TEMPORAL) {
+      scaleTypes = ["bin-linear", "time", "utc", "ordinal", "bin-ordinal", "point", "band"];
```
kanitw: ["time", "utc"]

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -63,81 +92,217 @@ const POSITION_FIELD_QUANTITATIVE_INDEX = {
...
   } else {
     throw new Error('Property combination not recognized');
   }
 }
 
 
+// TODO: Eventually refactor to Vega-Lite
+function getSupportedScaleTypes(shelfId: ShelfId, fieldDef: ShelfFieldDef): string[] {
+  let scaleTypes: string[] = [];
+  if (contains([Channel.X, Channel.Y], shelfId.channel)) {
```
kanitw: Better check switch case by data type first, then do if else for channel inside.  The method will be way less redundant.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -181,9 +352,42 @@ export function generateFormData(shelfId: ShelfId, fieldDef: ShelfFieldDef) {
       const prop = customProp.prop;
       const nestedProp = customProp.nestedProp;
       const propertyKey = nestedProp ? prop + '_' + nestedProp : prop;
-      formData[propertyKey] = fieldDef[prop] ? nestedProp ? fieldDef[prop][nestedProp] : fieldDef[prop] : undefined;
+      const fData = fieldDef[prop] ? nestedProp ? fieldDef[prop][nestedProp] : fieldDef[prop] : undefined;
+      // Display empty string when '?' is passed in to retrieve default value
+      // '?' is passed when formData is empty to avoid passing in empty string as a property/nestedProp value
+      formData[propertyKey] = fData === '?' ? '' : fData;
```
kanitw: Can you point me to a place where '?' is passed in?  This seems a bit brittle.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -63,81 +92,217 @@ const POSITION_FIELD_QUANTITATIVE_INDEX = {
...
       const schemaProperty: StringSchema = {
-        title: title,
-        enum: (vlSchema as any).definitions.ScaleType.enum,
-        type: 'string'
+        ...baseUiSchema,
+        enum: isContinuous(fieldDef) ? SEQUENTIAL_COLOR_SCHEMES : CATEGORICAL_COLOR_SCHEMES,
       };
       return generateSchema(prop, nestedProp, DEFAULT_SELECT_UISCHEMA, propTab, schemaProperty);
+    } else if (nestedProp === 'range' || nestedProp === 'domain') {
+      const schemaProperty: StringSchema = {
```
kanitw: 1) why do you need to do ... at all, can't you just use baseUISchema
2) baseUISchema, not baseUiSchema
3) I just notice this is a property's schema, not a schema's property.   Better name all of them `schemaProperty`?

src/components/encoding-pane/property-editor-schema.ts
ssharif6: https://github.com/vega/voyager/pull/799/files#diff-f09c2cf3fa6f34324d8cf3c87e68f796R72

This is used as a way to protect against users passing in blank data for text input form values. So if a user types in a domain, and then deletes all the values, it doesn't break Voyager.

src/components/encoding-pane/property-editor-schema.ts
ssharif6: Done! https://github.com/vega/vega-lite/issues/3725

src/components/encoding-pane/property-editor.tsx
```diff
@@ -67,7 +68,33 @@ export class PropertyEditorBase extends React.PureComponent<PropertyEditorProps,
...
+          throw new Error('Invalid format for domain. Must follow format: Min Number, Max Number');
+        } else if (fieldDef.type === ExpandedType.TEMPORAL) {
+          // TODO: Not supported yet
+          throw new Error('Voyager does not currently support temporal domain values');
+        }
+        return result === '' ? '?' : domain;
+      }
+    }
+
+    // if form data is empty, default to auto suggested view, which is ? in compass
```
kanitw: I think this logic won't work because Compass doesn't recommend properties of scale, axis, legend.   The right thing to is do set to `undefined` and it will use Vega-Lite's default.  

Thus, you should revise all logic that deals with this `'?'`.

src/components/encoding-pane/property-editor-schema.ts
kanitw: Thanks -- see https://github.com/vega/voyager/pull/799/files#r187163447

PR #800, opened 2018-05-01
[Feature Branch] Custom Encoding

kanitw:


kanitw:
@ssharif6  One critical problem found:

- [ ] By default `axis.title` is `undefined`, meaning we use default title.  However, if we set axis title to customize text and change our mind (wanting to use default title), it's impossible to do that anymore since removing all the text will set the title to an empty string.

kanitw:
Also, I still see broken layout 

image

I wanna see how it's like on your computer to check if it's just something wrong on my computer. (I'm doubtful about that though.)

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -0,0 +1,328 @@
...
+    type: 'string'
+  };
+
+  const {channel} = shelfId;
+
+  if (prop === 'scale') {
+    if (nestedProp === 'type') {
+      // Filtering based on channel type & fieldDef type
+      const scaleTypes: string[] = getSupportedScaleTypes(shelfId, fieldDef);
+      const propertySchema = {
```
kanitw: @ssharif6 The need to declare this propertySchema and call generateSchema with so many parameters is every time makes this method a bit painful to read. 

We should 
1) break this method to multiple method (`generateScale/Axis/LegendEditorSchema`)

2) use `switch-case` instead of `if`  since these are all exact match comparison

3) just pass `propertyKey` -- we don't really need `prop: string, nestedProp: string` inside?

4) Make specialized forms of generateSchema. 

- For example, enum is always used with select box.  So you should instead have 
`generateSelectSchema(propKey, title, enum)`. 

- Similarly, `generateTextBoxSchema(propKey, title, 'string' | 'number')`. 


So there are way less parameters and easier to read / maintain.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -232,27 +223,52 @@ function getSupportedScaleTypes(shelfId: ShelfId, fieldDef: ShelfFieldDef): stri
...
 }
 
-/*
-*  NOTE: factory method where propertyKey follows naming convention: prop_nestedProp
-*  Example: {prop: axis, nestedProp: orient} translates to "axis_orient"
-*/
-function generateSchema(prop: string, nestedProp: string, uiSchemaItem: UISchemaItem, propTab: string,
-                        schemaProp: SchemaProperty): PropertyEditorSchema {
-  const propertyKey = nestedProp ? prop + '_' + nestedProp : prop;
+function generateSelectSchema(propKey: string, enumVals: string[], baseUISchema: SchemaProperty) {
```
kanitw: Not sure why do you need `baseUISchema` as parameter for each of these method esp since they are all the same.   

(You might not notice that I intentionally left it from the method signature I suggested earlier.) 

Better extract it to be a constant BASE_UI_SCHEMA at the top-level and stop passing it around.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -232,27 +223,52 @@ function getSupportedScaleTypes(shelfId: ShelfId, fieldDef: ShelfFieldDef): stri
...
+  const schemaProp = {
```
kanitw: Also no need to declare this `schemaProp` thing ... given it is only used once and has a bit confusing name anyway.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -137,73 +114,83 @@ export const SEQUENTIAL_COLOR_SCHEMES = ['blues', 'greens', 'greys', 'purples',
...
+    case 'title':
+      return generateTextBoxSchema(propertyKey, '', baseUISchema, 'string');
+    default:
+      throw new Error('Property combination not recognized');
   }
 }
 
+function generateScaleEditorSchema(nestedProp: string, scaleTypes: string[], fieldDef: ShelfFieldDef,
+                                   baseUISchema: SchemaProperty) {
+  const propertyKey = nestedProp ? 'scale_' + nestedProp : 'scale';
```
kanitw: 1) when would there be no nestedProp for scale?

2) Also, why do some methods have propertyKey in the signature, but some do not? Please be consistent. 

If you haven't, always look at all of your diff via Source Tree or equivalent tools before committing.  This will improve your code quality. :)

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -137,73 +114,83 @@ export const SEQUENTIAL_COLOR_SCHEMES = ['blues', 'greens', 'greys', 'purples',
...
+    case 'legend':
+      return generateLegendEditorSchema(nestedProp, baseUISchema, propertyKey);
+    case 'format':
+      return generateFormatEditorSchema(baseUISchema);
+    default:
+      throw new Error('Property combination not recognized');
+  }
+}
+
+function generateFormatEditorSchema(baseUISchema: SchemaProperty) {
```
kanitw: For non-nested ones, these are one-liners anyway -- you can just return above -- avoid unnecessary code jump.  

(For nested ones, code jump is more justified as switch within switch is very hard to read.)

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -137,73 +114,83 @@ export const SEQUENTIAL_COLOR_SCHEMES = ['blues', 'greens', 'greys', 'purples',
...
-      return generateSchema(prop, nestedProp, DEFAULT_SELECT_UISCHEMA, propTab, propertySchema);
-    }
-  } else if (prop === 'format') {
-    return generateSchema(prop, nestedProp, DEFAULT_TEXT_UISCHEMA, propTab, baseUISchema);
-  } else {
-    throw new Error('Property combination not recognized');
+      return generateScaleEditorSchema(nestedProp, scaleTypes, fieldDef, baseUISchema);
+    case 'axis':
+      return generateAxisEditorSchema(nestedProp, channel as Channel, baseUISchema, propertyKey);
+    case 'stack':
```
kanitw: Can you add line breaks between all cases?

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -137,73 +114,83 @@ export const SEQUENTIAL_COLOR_SCHEMES = ['blues', 'greens', 'greys', 'purples',
...
-      };
-      return generateSchema(prop, nestedProp, DEFAULT_SELECT_UISCHEMA, propTab, propertySchema);
-    }
-  } else if (prop === 'format') {
-    return generateSchema(prop, nestedProp, DEFAULT_TEXT_UISCHEMA, propTab, baseUISchema);
-  } else {
-    throw new Error('Property combination not recognized');
+      return generateScaleEditorSchema(nestedProp, scaleTypes, fieldDef, baseUISchema);
+    case 'axis':
+      return generateAxisEditorSchema(nestedProp, channel as Channel, baseUISchema, propertyKey);
```
kanitw: why casting?

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -137,73 +114,83 @@ export const SEQUENTIAL_COLOR_SCHEMES = ['blues', 'greens', 'greys', 'purples',
...
+    default:
+      throw new Error('Property combination not recognized');
+  }
+}
+
+function generateFormatEditorSchema(baseUISchema: SchemaProperty) {
+  return generateTextBoxSchema('format', '', baseUISchema, 'string');
+}
+
+function generateLegendEditorSchema(nestedProp: string, baseUISchema: SchemaProperty, propertyKey: string) {
```
kanitw: nestedProp => `legendProp: keyof Legend` - -same for scale/axis methods.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -232,27 +223,52 @@ function getSupportedScaleTypes(shelfId: ShelfId, fieldDef: ShelfFieldDef): stri
...
+  const schemaProp = {
+    ...baseUISchema,
+    enum: enumVals
+  };
+
   const schema: ObjectSchema = {
     type: 'object',
     properties: {
-      [propertyKey]: schemaProp
+      [propKey]: schemaProp
```
kanitw: If you gonna use propKey, use propKey the whole file.  (Or just `propertyKey` the whole file.)

src/components/encoding-pane/property-editor-schema.ts
ssharif6: Since `ShelfId` is a union type of `ShelfChannelId` and `ShelfWildcardChannelId`, and therefore `Channel` prop can be either of type `Channel` or `SHORT_WILDCARD|Wildcard<Channel>`. 

I'm not too familiar with union typing in typescript, so I figured you'd have to eventually cast it to fit the Channel type, since we are not currently supporting wildcards (ShelfWildcardChannelId)

ssharif6: This issue arises quite a lot for me, as there are union types but my methods requires a specific type. How do I get around this in typescript? I know there is typeguards, but if you call the typeguard and it returns true, don't you still have to eventually cast it to that type?

src/components/encoding-pane/property-editor-schema.ts
ssharif6: This makes sense. When I call this method (and the other subsequent methods), how do I make it so that `nestedProp`, which is a string, is a `keyof Legend`, for instance. I had to cast, but I feel like that is not the right way to do it.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -115,78 +116,76 @@ export function generatePropertyEditorSchema(prop: string, nestedProp: string, p
...
-  const baseUISchema: SchemaProperty = {
-    title: title,
-    type: 'string'
-  };
   const {channel} = shelfId;
   switch (prop) {
     case 'scale':
       const scaleTypes: string[] = getSupportedScaleTypes(shelfId, fieldDef);
-      return generateScaleEditorSchema(nestedProp, scaleTypes, fieldDef, baseUISchema);
+      return generateScaleEditorSchema(nestedProp as keyof Scale, scaleTypes, fieldDef, title, propertyKey);
```
ssharif6: @kanitw 

How do I tell Typescript that nestedProp, which is a string, is a `keyof Scale`?

kanitw:
>  since we are not currently supporting wildcards (ShelfWildcardChannelId)

If we don't support `ShelfWildcardChannelId`, why does this component accept shelfId at all? 
This component should then accept only `channel` so it is only used with shelfId that's a channel?

kanitw:
> How do I tell Typescript that nestedProp, which is a string, is a keyof Scale?

Yeah, that would require some type guards, but definitely too much typing.  Since this casting is always correct, let's say it's ok here. 

(The casting for channel is not always correct so that's a bit more problematic.) 

Btw, testing is somehow still failing.

src/components/encoding-pane/property-editor.tsx
```diff
@@ -27,21 +28,24 @@ export class PropertyEditorBase extends React.PureComponent<PropertyEditorProps,
...
-          uiSchema={uiSchema}
-          formData={formData}
-          onChange={this.changeFieldProperty}
-        >
-          <button type="submit" style={{display: 'none'}}>Submit</button>
-          {/* hide required submit button */}
-        </Form>
-      </div>
-    );
+    if (!isWildcardChannelId(shelfId)) {
```
kanitw: `const {channel} = shelfId;` then apply type guard to channel so you don't have to cast.  

(As said, avoid casting at all cost unless it's really super annoying.)

PR #805, opened 2018-05-11
Add css property when wildcards are disabled

### LINKED ISSUE #784: Field pane should fill the space when we hide wildcards

abanh206:
Fix #784

### DIFF
--- a/src/components/data-pane/field-list.scss
+++ b/src/components/data-pane/field-list.scss
@@ -7,7 +7,11 @@
   margin-bottom: 4px;
 }
 
+/*
+Subtract 159px for the size of header and other objects in the field-pane
+Subtract 8px for padding between the footer and the download logs
+*/
 .field-list-no-wildcards {
-  max-height: calc(75vh);
+  max-height: calc(100vh - 167px);
   overflow-y: auto;
 }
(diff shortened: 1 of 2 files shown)


abanh206:
Before: 

image

After:

image

abanh206:
Before: 

image

After:

image

src/components/data-pane/field-list.scss
```diff
@@ -6,3 +6,8 @@
 .field-list-item {
   margin-bottom: 4px;
 }
+
+.field-list-no-wildcards {
+  max-height: calc(100vh - 167px);
```
kanitw: Why `-167px`?  

Note that 167 is a prime number so I suspect that the allocated space for the debug pane would not be proportional to other size of the UI.   If there is a good reason, this is fine.  But I wanna understand how you get this number, which seems a bit arbitrary.

abanh206: 159px is the total of the height, padding and margins of the header and other items in the data-pane. Then I added 8px for padding as that seems to be the convention. I can change the padding if you would like to a different number. The reason the number is odd and prime is because the header is 33px tall.

kanitw: This is fine, but better add a comment describing where 167px comes from.

PR #806, opened 2018-05-15
Filter customizers only for channels that support customization

### LINKED ISSUE #804: Expanding detail channel shows empty popup
- Better add a customizer for text's `format`
- Better not show caret for detail and make it unclickable

ssharif6:
Fix #804 

Disable customizer popup caret option for Detail channel. 
Created constant for channels that support encoding customization.

Brief demo showing customizer doesn't display caret for channels that have no customization supported yet. 

detailchannel

### DIFF
--- a/src/components/encoding-pane/encoding-shelf.tsx
+++ b/src/components/encoding-pane/encoding-shelf.tsx
@@ -19,7 +19,7 @@ import {DraggedFieldIdentifier, Field} from '../field/index';
 import * as styles from './encoding-shelf.scss';
 import {FieldCustomizer} from './field-customizer';
 import {FunctionPicker, FunctionPickerWildcardHandler} from './function-picker';
-import { CUSTOMIZABLE_ENCODING_CHANNELS } from './property-editor-schema';
+import {CUSTOMIZABLE_ENCODING_CHANNELS} from './property-editor-schema';
 
 /**
  * Props for react-dnd of EncodingShelf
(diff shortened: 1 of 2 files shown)


src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -18,6 +19,7 @@ import {DraggedFieldIdentifier, Field} from '../field/index';
 import * as styles from './encoding-shelf.scss';
 import {FieldCustomizer} from './field-customizer';
 import {FunctionPicker, FunctionPickerWildcardHandler} from './function-picker';
+import { CUSTOMIZABLE_ENCODING_CHANNELS } from './property-editor-schema';
```
kanitw: Please do not include space between braces.  If we don't have this a part of linter yet, please submit a separate PR to enforce this.

src/components/encoding-pane/property-editor-schema.ts
```diff
@@ -58,6 +58,9 @@ export interface PropertyEditorSchema {
   schema: ObjectSchema;
 }
 
+// Currently supported customizble encoding channels that display caret in customizer UI
+export const CUSTOMIZABLE_ENCODING_CHANNELS = [Channel.X, Channel.Y, Channel.COLOR, Channel.SIZE, Channel.SHAPE];
```
kanitw: How about row and column?

src/components/encoding-pane/encoding-shelf.tsx
ssharif6: Sorry every time I use VS Code this happens. Will make a PR to add this rule to linter.

src/components/encoding-pane/property-editor-schema.ts
ssharif6: Those are not supported as of now. Eventually it'd be nice to customize the row/column encoding channels. It'd be nice to support customizing headers and whatnot. 

So better just leave what is working right now, and then when add support for row/column, update this list as well

PR #811, opened 2018-05-16
Fix the filter range bug

### LINKED ISSUE #708: Cannot type max number in range filter
We shouldn't use onChange event listener

abanh206:
Fix #708


kanitw:
Not sure why early return solves the problem.  

Can you (1) provide a GIF demonstrating that it solves the problem and (2) give more explanation why?

abanh206:
An early return solves the problem because it stops before the function fires the filter modify action.

[
filterrangefix
](url)

kanitw:
I don't see how this GIF is demonstrating that you are fixing #708.  

>  if the min is 8, the max is 22, we cannot type 25.

Try **typing** the number.

kanitw:
I'll close this PR for now since it's probably not the right fix.  Let's open another one when you have the right fix ready :)

Issue #662, opened 2017-08-24
UI for customizing constant value (for valuedef) 


kanitw:
- [ ] Constant Value for non-position only
  - We need to make values `SpecificEncoding` be `ShelfFieldDef | ValueDef`
  - This will involve some refactor throughout the projects as we need to add some type guards and some type checking.

ZeningQu:
Just noticed [this vega example](https://vega.github.io/editor/#/examples/vega/distortion-comparison) picks color like below. 

![vega pick color example](https://user-images.githubusercontent.com/822034/36884749-9ccc06f6-1d97-11e8-96e7-3f23a4546dd8.gif)

Not saying this is the way to go for Voyager, but I want to mark this example.

kanitw:
Thanks, mac has native color picker that works pretty well.  I don't know how it works in Linux / windows though.  If there is a good standard, then we shouldn't reinvent the wheel. :)

PR #816, opened 2018-05-21
Add ValueDef Model & Reducer

ssharif6:
Add `ShelfValueDef`, `SPEC_VALUE_CHANGE` action & reducer. 

The PR following this PR will address #662 , which will use existing infrastructure from Field-Customizer and infrastructure from this PR


kanitw:
What's the difference between this and https://github.com/vega/voyager/pull/664?

ssharif6:
It does not include the component for the field customizer, and it is updated on current version of master + some minor differences. Wasn't sure if we wanted to do a feature branch and merge into that branch like we did for CustomEncoding, since Compass isn't supporting this yet and we still need to upgrade CQL and VL in voyager. If you want, I can just edit the existing PR #664.

kanitw:
If Compass isn't supporting this yet, we shouldn't merge it yet.  For the other PR, if the other one is not needed, just close it and says we have this one.

PR #827, opened 2018-07-23
WIP -- Value Def Customization UI

ssharif6:
UI for customization of constant values (valuedef). Addresses issue #662.

### DIFF
--- a/src/queries/alternative-encodings.test.ts
+++ b/src/queries/alternative-encodings.test.ts
@@ -58,4 +58,57 @@ describe('queries/alternative-encodings', () => {
       chooseBy: ['aggregationQuality', 'effectiveness']
     });
   });
+
+  it.only('should correctly produce a query given ValueDef encoding', () => {
+    const query: Query = {
+      spec: {
+        transform: [{
+          filter: {
+            field: 'a',
+            oneOf: ['1, 2']
+          }
+        }],
+        mark: 'point',
+        encodings: [{
+          channel: 'y',
+          field: 'a',
+          type: 'quantitative'
+        }, {
+          channel: 'color',
+          value: '#ffffff'
+        }, {
+          channel: 'x',
+          field: 'c',
+          type: 'temporal'
+        }]
+      }
+    };
+
+    expect(alternativeEncodings.createQuery(query)).toEqual({
+      spec: {
+        transform: [{
+          filter: {
+            field: 'a',
+            oneOf: ['1, 2']
+          }
+        }],
+        mark: SHORT_WILDCARD,
+        encodings: [{
+          channel: SHORT_WILDCARD,
+          field: 'a',
+          type: 'quantitative'
+        }, {
+          channel: 'color',
+          value: '#ffffff'
+        }, {
+          channel: SHORT_WILDCARD,
+          field: 'c',
+          type: 'temporal'
+        }]
+      },
+      groupBy: 'encoding',
+      orderBy: ['fieldOrder', 'aggregationQuality', 'effectiveness'],
+      chooseBy: ['aggregationQuality', 'effectiveness']
+    });
+  });
 });


src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -5,6 +5,7 @@ import * as CSSModules from 'react-css-modules';
 import {ConnectDropTarget, DropTarget, DropTargetCollector, DropTargetSpec} from 'react-dnd';
 import * as TetherComponent from 'react-tether';
 import {contains} from 'vega-lite/build/src/util';
+import {isValueDef} from '../../../node_modules/vega-lite/build/src/fielddef';
```
kanitw: Remove `../../../node_modules/`.  Please also check if you have accidentally included `node_modules` like this in other places too.  :)

src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -91,27 +93,35 @@ class EncodingShelfBase extends React.PureComponent<
             attachment="top left"
             targetAttachment="bottom left"
           >
-            {(fieldDef && !isWildcardChannelId(id) && contains(CUSTOMIZABLE_ENCODING_CHANNELS, id.channel)) ?
+            {(!isWildcardChannelId(id) && contains(CUSTOMIZABLE_ENCODING_CHANNELS, id.channel)) ?
```
kanitw: Since we don't allow customizing x and y valuedef, caret shouldn't be shown for x and y if there is no fieldDef.

kanitw:
- [ ] There is a bug in CompassQL for size value.  (Putting quantitative X normally get ticks, but adding size value change ticks to points.) 


Similarly, bars got changed to points too: 

image

And line got changed to area:

image


In addition to fixing these, perhaps, it's worth trying more different combination of theses.

- [ ] The Value def UI is not styled yet. Better use consistent styles with fieldDef customizer.  For example, text could look like axis title. 

image

- [ ] `shape` value customizer should be enabled only if the mark is point, `text` value customizer should be enabled only if the mark is text

- [ ] For all of these values, there is no way to set "use default" value. 

- [ ] For text, if I type something and then later delete the text, then I get weird empty fieldDef.

image



- [ ] Size of point, text, and bar use different units (pixel area, font pixel, and width pixel -- we have to think about how we gonna make size work without causing problem to other marks.

src/components/encoding-pane/value-customizer.tsx
```diff
@@ -0,0 +1,56 @@
...
+          onChange={this.changeValue}
+        >
+          <button type="submit" style={{display: 'none'}}>Submit</button>
+        </Form>
+      </div>
+    );
+  }
+
+  protected changeValue(result: any) {
+    const value = result.formData[Object.keys(result.formData)[0]].toString();
```
kanitw: why `toString()`.  Isn't size not a string, for example?

src/queries/alternative-encodings.ts
```diff
@@ -1,6 +1,7 @@
 
 import {Query} from 'compassql/build/src/query/query';
 import {isWildcard, SHORT_WILDCARD} from 'compassql/build/src/wildcard';
+import {isValueQuery} from '../../node_modules/compassql/build/src/query/encoding';
```
kanitw: node_modules

src/components/encoding-pane/encoding-shelf.tsx
```diff
@@ -91,27 +93,35 @@ class EncodingShelfBase extends React.PureComponent<
...
             {this.state.customizerIsOpened &&
-            <div ref={this.popupRefHandler}>
-              <FieldCustomizer
-                shelfId={id}
-                fieldDef={fieldDef}
-                handleAction={handleAction}
-              />
-            </div>
+              <div ref={this.popupRefHandler}>
+                { (!fieldDef && !valueDef) || isValueDef(valueDef) ?
```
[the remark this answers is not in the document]
ssharif6: Fixed! The issue was here: https://github.com/vega/voyager/blob/master/src/components/encoding-pane/index.tsx#L117

I fixed this by adding appropriate typeguarding and the basic `!fieldDef` logic can now work correctly

PR #848, opened 2020-06-15
timestamp corrected in README

researcherben:
timestamp in readme did not link to correct location in video. Fixed URL to use seconds


domoritz:
Thank you!

PR #849, opened 2020-06-16
New Dockerfile

tjstavenger-pnnl:
* Add Dockerfile to run datavoyager (would be nice to get this built and deployed to https://hub.docker.com/)
* Configured server to listen on 0.0.0.0 to open server up for Docker -- this has an artifact of opening other deployements up to all IPs, too
* Added yarn start:headless option to remove the `--open` parameter as it isn't needed in Docker
* Updated README.md with Docker build & run instructions

While running Travis CI build on my fork to verify my changes didn't break anything tested, I found the build has an out of memory error. I increased the `--max-old-space-size` to work around the issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM node:9.11.2
+
+COPY . /voyager
+
+RUN cd voyager && yarn && yarn build
+
+WORKDIR /voyager
+
+EXPOSE 9000
+
+CMD ["yarn", "start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:9.11.2
+FROM node:6.17.1
 
 COPY . /voyager
 
@@ -8,4 +8,4 @@ WORKDIR /voyager
 
 EXPOSE 9000
 
-CMD ["yarn", "start"]
+CMD ["yarn", "start:headless"]
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ Build Docker image
 docker build -t vega/voyager .
 ```
 
-Run Docker container
+Run Docker container. The `-p 9000:9000` parameter is required to publish the container's yarn port to a host port, allowing users to view the datavoyager tool at http://localhost:9000/.
 ```
 docker run -p 9000:9000 vega/voyager
 ```
--- a/README.md
+++ b/README.md
@@ -156,3 +156,16 @@ This will run Voyager in "server-mode" sending requests to voyager-server, which
 The server url is controlled by the `SERVER` environment variable.
 
 See [voyager-server](https://github.com/vega/voyager-server) for more information on what portions of the functionality the server handles.
+
+## Docker
+Docker image based on [node:9.11.2](https://hub.docker.com/layers/node/library/node/9.11.2/images/sha256-571a95847b472f5a08c2287c3f36cfa4bb88c483fa5866504dc7cc8e3abfb588?context=explore)
+
+Build Docker image
+```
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:6.17.1
```
domoritz, line 1: Can you use node 12 or 14? Node 6 is not maintained anymore.

README.md
```diff
@@ -156,3 +156,21 @@ This will run Voyager in "server-mode" sending requests to voyager-server, which
 The server url is controlled by the `SERVER` environment variable.
 
 See [voyager-server](https://github.com/vega/voyager-server) for more information on what portions of the functionality the server handles.
+
+## Docker
+Docker image based on [node:6.17.1](https://hub.docker.com/layers/node/library/node/6.17.1/images/sha256-0bb1f1235ffb537f2543de5fa0de7e4e93f522dd9cc6da51da255cffbdd99aa3?context=explore)
```
domoritz, line 161: Would it make sense to use VSCode's container development feature instead? https://code.visualstudio.com/docs/remote/containers

Dockerfile
tjstavenger-pnnl: I picked 6.17.1 as it is what the Travis CI build uses. The newer versions of Node are incompatible with vega/voyager -- at least the builds failed for me. It looks like the voyager code uses removed APIs (likely long past deprecated).

The knock/voyager Dockerfile I based this on had used node 9

README.md
tjstavenger-pnnl: I'm not familiar with VSCode's development container feature. It certainly looks interesting.

In my case, my deployment environment requires the Docker container, though I could potentially see the same container reused for development testing & debugging.

Dockerfile
domoritz: Maybe we should switch to github actions like we did in all other active vega projects.

README.md
domoritz: Okay. I'll let @kanitw make the call whether he wants a docker container in master right now. Even if not, it'll be great to have this pull request as a reference if someone wants to use docker.

Dockerfile
tjstavenger-pnnl: Certainly an option, but I think you'd still need to build with the older version of node until the voyager software is updated. The .travis.yml defines the version of node used, not Travis itself. https://github.com/vega/voyager/blob/master/.travis.yml#L3

tjstavenger-pnnl: A quick change to the Dockerfile to use node:latest (presumably 14.4) and it also looks like there are library version incompatibilities. I had tried earlier today to force upgrading all npm libraries to current/latest, but that was unsuccessful too. I hadn't spent the time to dive deeper into the software to determine what versions of each of the dependencies are required. I had really just wanted to get it working, so I fell back to the known working & tested version of 6.17.1.

domoritz: Since node 12 is the oldest supported version, let's hold off on merging this pull request. 

https://nodejs.org/en/about/releases/

tjstavenger-pnnl:
I brought in the same changes from PR #839 to updated node-sass. That lets it build with Node 14.4.0. I also changed the .travis-ci.yml and Dockerfile to use 14.4.0

I didn't change any of the other dependencies -- quite a few are out of date. I took a couple of hours yesterday to try to just update all to the latest and quickly found there are code and dependency incompatibilities. I made it through a few of them, but decided to stop when I got pretty deep into the code. I figure it would be better left to a person more familiar with the code base.

PR #858, opened 2021-04-14
Upgrade 30 dependencies to the latest version (Apr. 21)

joos2010kj:
List of yarn-upgraded dependencies:

- case-sensitive-paths-webpack-plugin
- font-awesome-sass-loader
- fs-extra
- gh-pages
- isomorphic-fetch
- moment
- node-sass
- normalize-scss
- papaparse
- react-addons-test-utils
- react-copy-to-clipboard
- react-css-modules
- react-file-download
- react-tabs
- react-tabs-redux
- re-reselect
- style-loader
- tether
- throttle-debounce
- tslib
- webpack-notifier
- @types/classnames
- @types/d3-selection
- @types/fetch-mock
- @types/isomorphic-fetch
- @types/json-stable-stringify
- @types/papaparse
- @types/react-css-modules
- @types/throttle-debounce
- @types/webpack-env


package.json
```diff
@@ -38,61 +38,63 @@
   },
   "dependencies": {
     "ajv": "^5.2.2",
+    "codecov": "^3.8.1",
```
domoritz, line 41: Please remove

package.json
```diff
@@ -38,61 +38,63 @@
...
     "compassql": "0.20.2",
     "font-awesome": "^4.7.0",
-    "font-awesome-sass-loader": "^1.0.3",
-    "isomorphic-fetch": "^2.2.1",
-    "moment": "^2.18.1",
-    "normalize-scss": "^6.0.0",
-    "papaparse": "^4.3.5",
+    "font-awesome-sass-loader": "^2.0.1",
```
domoritz, line 44: This should be a dev dependency I think

package.json
```diff
@@ -38,61 +38,63 @@
...
+    "isomorphic-fetch": "^3.0.0",
+    "moment": "^2.29.1",
+    "normalize-scss": "^7.0.1",
```
domoritz, line 47: This should be a dev dependency I think

package.json
```diff
@@ -38,61 +38,63 @@
...
-    "vega-datasets": "^1.8.0",
+    "vega-datasets": "^2.2.0",
     "vega-expression": "^2.3.1",
     "vega-lite": "2.4.0",
-    "vega-typings": "^0.2.15",
     "vega-tooltip": "^0.4.2",
-    "vega-util": "1.7.0"
+    "vega-typings": "^0.2.15",
+    "vega-util": "1.7.0",
+    "webpack-notifier": "^1.13.0"
```
domoritz, line 83: Why was this added?

package.json
```diff
@@ -38,61 +38,63 @@
...
     "compassql": "0.20.2",
     "font-awesome": "^4.7.0",
-    "font-awesome-sass-loader": "^1.0.3",
-    "isomorphic-fetch": "^2.2.1",
-    "moment": "^2.18.1",
-    "normalize-scss": "^6.0.0",
-    "papaparse": "^4.3.5",
+    "font-awesome-sass-loader": "^2.0.1",
+    "isomorphic-fetch": "^3.0.0",
+    "moment": "^2.29.1",
```
domoritz, line 46: Can we get rid of this? Moment was useful a few years ago but there are lighter alternatives today.

domoritz (changes requested):

domoritz:
Let me know when this is ready. For now, I'll make it as draft.

package.json
joos2010kj: What should I replace this moment with? It is marked as a peer dependency of react-moment.

domoritz: Leave it for now then but we may not need react moment anymore. We can just use the browser APIs for time.

domoritz:
Remove the yarn error log file.

joos2010kj:
The font issue has been resolved.

src/components/app.scss
```diff
@@ -1,3 +1,5 @@
+$base-font-family: sans-serif !default;
```
domoritz, line 1: Can you remove the !default and move this after the normalize import?

joos2010kj: Hi,

I have tried to put it AFTER @import initially, but the font change did not kick in.

So I googled a bit, and tried it similar to what this stackoverflow post said (https://stackoverflow.com/questions/19793482/how-can-i-set-basefontfamily-for-bootstrap) by putting BEFORE it, and it worked.

domoritz:
Thank you!

